Issue #17, opened 2016-11-03
Model Sharing


virajprabhu:
We want to have a feature to generate a shareable link to a network built by a user that another user can view, edit. We will need to store the link and prototxt in our database when the user clicks share, and look up and regenerate the network from DB when another user tries to access the link.

Eventually we want to also allow collaborative editing of networks with real time updates.

@deshraj @dexter1691.

souravsingh:
I would like to work on the issue. How do I get started?

deshraj:
@souravsingh sorry for replying late. I missed this thread somehow. You can start with the integration of django-allauth and then after setting up the authentication, we can work on generating the sharable link as a new feature. Does that sound good?

souravsingh:
@deshraj Thanks, I will start working on it.

jainamritanshu:
@deshraj @virajprabhu Can I work on this issue if it's still open and not assigned to anyone else?

virajprabhu:
@jainamritanshu This has not been assigned yet, but I'd recommend you take on one issue at a time so as to not block others. I'd recommend starting with #14 that you expressed interest in, since that is a simpler issue to get started with.

utsavgarg:
@virajprabhu I have added support for static model sharing, the flow of events is as follows:

- export prototxt to url, a unique id will be given and the prototxt will be saved on the server
- import prototxt from url, enter the generated id, and the prototxt will load

![output_e6v7ot](https://cloud.githubusercontent.com/assets/11680179/24166551/9f242c48-0e99-11e7-9838-3fab567678e1.gif)

PR #48, opened 2017-03-22
added static sharing

utsavgarg:
Added support for static sharing as discussed in #17

### DIFF
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -4,7 +4,6 @@ from django.http import JsonResponse
 from django.conf import settings
 import yaml
 import os
-#BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 from caffe.proto import caffe_pb2
 from google.protobuf import text_format
 
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -1,9 +1,10 @@
 from django.shortcuts import render
 from django.views.decorators.csrf import csrf_exempt
 from django.http import JsonResponse
+from django.conf import settings
 import yaml
 import os
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+#BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 from caffe.proto import caffe_pb2
 from google.protobuf import text_format
 
@@ -17,7 +18,7 @@ def importPrototxt(request):
                 return JsonResponse({'result': 'error', 'error': 'No Prototxt model file found'})
         elif 'proto_id' in request.POST:
             try:
-                prototxt = open(os.path.join(BASE_DIR,'media',request.POST['proto_id']+'.prototxt'), 'r')
+                prototxt = open(os.path.join(settings.BASE_DIR,'media',request.POST['proto_id']+'.prototxt'), 'r')
             except Exception:
                 return JsonResponse({'result': 'error', 'error': 'No Prototxt model file found'})
         caffe_net = caffe_pb2.NetParameter()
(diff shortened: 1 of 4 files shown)


caffe_app/views/import_prototxt.py
```diff
@@ -2,25 +2,31 @@
...
-        try:
-            prototxt = request.FILES['file']
-        except Exception:
-            return JsonResponse({'result': 'error', 'error': 'No Prototxt model file found'})
-
+        if 'file' in request.FILES:
+            try:
+                prototxt = request.FILES['file']
+            except Exception:
+                return JsonResponse({'result': 'error', 'error': 'No Prototxt model file found'})
```
deshraj: Can you please change the response code to `{'error': 'No Prototxt model file found'}` in case of any error. The current key value pair `'result': 'error'` is just duplication.

caffe_app/views/import_prototxt.py
```diff
@@ -2,25 +2,31 @@
...
+        elif 'proto_id' in request.POST:
+            try:
+                prototxt = open(BASE_DIR+'/media/'+request.POST['proto_id']+'.prototxt', 'r')
```
deshraj: Can you please use `os,path.join()` in this line?

caffe_app/views/import_prototxt.py
```diff
@@ -2,25 +2,31 @@
...
+            except Exception:
+                return JsonResponse({'result': 'error', 'error': 'No Prototxt model file found'})
```
deshraj: Same comment as above. :)

caffe_app/views/import_prototxt.py
```diff
@@ -2,25 +2,31 @@
...
 
 @csrf_exempt
 def importPrototxt(request):
     if request.method == 'POST':
-        try:
-            prototxt = request.FILES['file']
-        except Exception:
-            return JsonResponse({'result': 'error', 'error': 'No Prototxt model file found'})
-
+        if 'file' in request.FILES:
```
deshraj: Also, can you please add a check to limit the mimetype allowed at the backend for the uploaded files?

caffe_app/views/import_prototxt.py
utsavgarg: It is not a duplication, in content.js the condition checks if the result is a success or error and then shows the message depending on the situation.

utsavgarg: If I remove the `result': 'error'` then I'll have to add condition for both when the prototxt is not found and when it is invalid in `content.js`. I feel the way it is, the logic is much clearer.

utsavgarg:
@deshraj leaving the `'result': 'error'` issue, I have made the other changes.

caffe_app/views/import_prototxt.py
```diff
@@ -2,25 +2,30 @@
 from django.views.decorators.csrf import csrf_exempt
 from django.http import JsonResponse
 import yaml
+import os
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
deshraj: There is no need to declare the BASE_DIR again. You can just import it  

`from django.conf import settings`
And then use it simply as `settings.BASE_DIR`. 

For more information about this, please see https://docs.djangoproject.com/en/1.11/topics/settings/#using-settings-in-python-code

utsavgarg: Oh I didn't know that, thanks! I have made the change.

caffe_app/views/import_prototxt.py
```diff
@@ -1,26 +1,32 @@
 from django.shortcuts import render
 from django.views.decorators.csrf import csrf_exempt
 from django.http import JsonResponse
+from django.conf import settings
 import yaml
+import os
+#BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
deshraj: @utsavgarg Please remove the comment. There is no need to comment this. We can always recover this. :)

deshraj:
LGTM. Merging this. :)

PR #56, opened 2017-05-09
added travis

utsavgarg:


.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: node_js
+node_js:
+  - "node"
+  - "7"
+  - "6"
+  - "5"
```
deshraj: Umm...I am not sure what does the line 4,5,6 do here? Can you please elaborate this? if these not required, then please remove them.

utsavgarg: They checks with older versions of node as well, line 3 will check with the latest stable release.

Issue #51, opened 2017-03-31
TensorFlow in requirements.


achillesrasquinha:
I don't think TensorFlow should be part of the requirements.txt. Users could install this separately (or a build script can be written for the same - e.g. a Makefile) since TensorFlow wasn't completely registered under pip (*pip installable*) until their initial version release. This can also help new users of Fabrik to opt for GPU-tensorflow (which requires NVIDIA CUDA and CuDNN installed).

utsavgarg:
fixed in #64

achillesrasquinha:
👍

PR #64, opened 2017-05-11
Split requirements into production and dev

utsavgarg:
- Adds a new script to install caffe and tensorflow if required #51 
- Splits the requirements depending on use case


deshraj:
Can we have a common.txt which will contain the requirements that are shared across dev and production environment ?

utsavgarg:
@deshraj I can change the prod one to common, we currently don't have any dependencies which are required in prod and not in dev

deshraj:
Sounds good.

utsavgarg:
@deshraj you can now merge the changes

PR #76, opened 2017-06-01
Made layer selection as a nested dropdown

utsavgarg:
* Commented unused cog icon in topbar
* Grouped layers with similar functionality to make a nested layer selection


utsavgarg:
@virajprabhu Yeah missed that part, then we'll have to keep those dots

PR #78, opened 2017-06-12
Added support for import/export of Python layer

utsavgarg:
Allows to import and export any custom Python layer


caffe_app/views/import_prototxt.py
```diff
@@ -348,6 +348,26 @@ def importPrototxt(request):
...
+                    params['loss_weight'] = layer.loss_weight[0]
+                if (not layer.bottom):
+                    params['endPoint'] = '1, 0'
+                elif ('loss' in layer.name.lower()):
+                    params['endPoint'] = '0, 1'
+                else:
+                    params['endPoint'] = '1, 1'
+                for param in params:
+                    if isinstance(params[param], list):
+                        params[param] = str(params[param])[1:-1]
```
virajprabhu: What does this slicing do?

caffe_app/views/import_prototxt.py
```diff
@@ -348,6 +348,26 @@ def importPrototxt(request):
...
                 params['legacy_version'] = layer.contrastive_loss_param.legacy_version
 
+            # ********** Python Layer **********
+            elif(layer.type == 'Python'):
+                if (layer.python_param.module):
+                    params['module'] = layer.python_param.module
+                if (layer.python_param.layer):
+                    params['layer'] = layer.python_param.layer
+                if (layer.python_param.param_str):
+                    params.update(eval(layer.python_param.param_str))
```
utsavgarg: `update` add all keys from that dictionary as keys to the `params` dictionary.

caffe_app/views/import_prototxt.py
utsavgarg: After converting the `array` to `str`, '[`1:-1]` will remove the brackets.

PR #84, opened 2017-06-24
Added file format restriction for upload

utsavgarg:
* Allows only '.prototxt' file for Caffe
* Allows only '.pbtxt' file for Tensorflow


virajprabhu:
LGTM

PR #87, opened 2017-06-30
Added unit tests for Python layer

utsavgarg:


tests/unit/ide/test_views.py
```diff
@@ -982,3 +1000,4 @@ def test_shapes(self):
         for layer in model.layers:
             self.assertEqual(list(layer.output_shape[::-1][:-1]),
                              net[layer.name]['shape']['output'])
+
```
deshraj: Can we get rid of this please?

Issue #30, opened 2017-02-11
Import all the models from Caffe ModelZoo into IDE


dexter1691:
We should try to successfully import all the models from Caffe ModelZoo to see if some layers are not working. 
Consequently fix or add the missing the layers into the IDE. 

https://github.com/BVLC/caffe/wiki/Model-Zoo

For GSOC participants:
This seems like a very good starting point to see how IDE works. Here is what you have to do
- Choose one of the models from the link above and create an issue for it. Cross reference it here for others to see.
- We will assign the issue to you
- If the model is working. Add a unit test for it so that any later changes to the projects ensure compatibility with those models. 
- If the model isn't working, try to discuss with the mentors and fix the issue! 

We are maintaining the list of models tested by contributors here: 
https://github.com/Cloud-CV/IDE/wiki/Models

anweshknayak:
https://gist.github.com/jimmie33/7ea9f8ac0da259866b854460f4526034

This is working for me. @dexter1691 

![screenshot from 2017-02-12 02 16 02](https://cloud.githubusercontent.com/assets/8139783/22857343/7757562e-f0c9-11e6-83f8-ffbceb7ce64e.png)
![screenshot from 2017-02-12 02 15 47](https://cloud.githubusercontent.com/assets/8139783/22857345/7ab13bc8-f0c9-11e6-82a5-0f1ef9a9414a.png)

dexter1691:
Nice! Let's maintain a list of the one's that working and the ones that are not working. GoogleNet was already part of the ones that we used to check. 

Let's keep trying others to find out the ones that are not working. I am sure there will be a few!

anweshknayak:
https://gist.github.com/jimgoo/0179e52305ca768a601f Not working.

kris-singh:
I woul like to work on this

virajprabhu:
@anweshknayak Are you sure we have front-end support for all the layers in the prototxt?

@kris-singh Great, can you start with simple classification models like AlexNet and VGGNet? Would be great if you could try out a ResNet (it won't work at the moment) and open an issue with the layers that we are missing. It would also help you a good overall idea of the current status of the project.

Ram81:
@virajprabhu  I am interested in this one can you guide me?

virajprabhu:
[...]

PR #92, opened 2017-07-20
Added sample models to front end

utsavgarg:
#30 

* Improved layout logic to generalise better
* Added models 

image


caffe_app/views/import_prototxt.py
```diff
@@ -23,6 +23,14 @@ def importPrototxt(request):
             except Exception:
                 return JsonResponse({'result': 'error',
                                      'error': 'No Prototxt model file found'})
+        elif 'sample_id' in request.POST:
```
virajprabhu: What is `sample_id`?

PR #103, opened 2017-09-15
Fix #30: Add All-CNN model to Fabrik/Zoo

### LINKED ISSUE #30: Import all the models from Caffe ModelZoo into IDE

Ram81:
@utsavgarg I tested All-CNN model from https://github.com/mateuszbuda/ALL-CNN it works.
allcnn

### DIFF
--- a/example/caffe/All_CNN.prototxt
+++ b/example/caffe/All_CNN.prototxt
@@ -1,0 +1,358 @@
+name: "ALL-CNN-C"
+layer {
+  name: "data"
+  type: "Data"
+  top: "data"
+  top: "label"
+  include {
+    phase: TRAIN
+  }
+  data_param {
+    # path to GCN preprocessed and ZCA whitened cifar-10 train data
+    source: "cifar-10_train_lmdb"
+    batch_size: 256
+    backend: LMDB
+  }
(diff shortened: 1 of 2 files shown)


utsavgarg:
@Ram81 To make it work, you will also need to add this model to the models drop-down list so that its accessible from the front-end.

Ram81:
@utsavgarg sure I will do that

Ram81:
@utsavgarg Done :+1:

Issue #105, opened 2017-09-16
Add models to the Fabrik Model Zoo


utsavgarg:
We want to make the Fabrik ModelZoo a rich collection of models from all streams of deep learning research (vision, speech, text) and want to add models from both frameworks that are currently well supported by this project (Caffe and Keras).

For people looking to get started with contributing to this project, this is a very good starting point. To add a model to the model zoo, please do the following:
- Add the model prototxt/json to the `example/<framwork>` folder depending on the model being from caffe or keras
- Add the entry to the models drop-down list on the front-end 
- Test if loading the model and exporting it to both or at least one framework is working fine

utsavgarg:
@melvyniandrag we currently have some issue with Keras depenencies on the deplyed version which is why you might be facing problems in loading some models. I am not quite sure about the wrapping idea, if you could come up with a mock-up we can then discuss it further.

melvyniandrag:
I guess this ties into the issue with the docker containers too? I guess I'll work on this in a container and then if no one has made one yet Ill submit the docker file too.

deshraj:
@melvyniandrag: Sounds good. This is super critical so it would be great if you can submit the pull request really soon.

manuhortet:
Hello, I'm trying to add some models to the Fabrik ModelZoo. I've added the prototxts to the right `example/framework` folder and did the right calls from `ide/static/bundle/bundle.js`

The added models do appear in Actions > Load from zoo, but when I click them, nothing happens. No error message appears, so I'm not pretty sure what's going wrong. Any help here?

AtharvaKhare:
@manuhortet I think you need to modify `ide/static/js/topBar.js`. `ide/static/bundle/bundle.js` is generated by webpack.

Also, are you keeping your models in `example/keras` or `example/caffe`?

manuhortet:
I did what you suggested but it doesn't seem to work either. I'm keeping my models in example/caffe, as I got them from the Caffe ModelZoo and are `.prototxt` files.

I found this comment on `netLayout.js`: 

```
// designed to work with googleNet.
// will work with all prototxts which follows our assumption
// but the UI may not be clean.
```

Maybe I'm trying to load invalid prototxt files? Btw thank you for your help Atharva!

utsavgarg:
@manuhortet if the prototxt files are invalid, an error should pop up. In case its not clear to add a model to the model zoo for Caffe you have to do the two following things:
[...]

PR #107, opened 2017-09-24
Fix: Add ZF Net model to FabrikZoo(#105)

prakharchoudhary:
@utsavgarg In reference to issue #105 I found ZF net to be a viable addition to the models. Hence, I am making this PR.

### DIFF
--- a/example/caffe/zf_net.prototxt
+++ b/example/caffe/zf_net.prototxt
@@ -1,0 +1,497 @@
+name: "ZF"
+layer {
+  name: 'input-data'
+  type: 'Python'
+  top: 'data'
+  top: 'im_info'
+  top: 'gt_boxes'
+  python_param {
+    module: 'roi_data_layer.layer'
+    layer: 'RoIDataLayer'
+    param_str: "'num_classes': 21"
+  }
+}
+
+#========= conv1-conv5 ============
+
+layer {
+	name: "conv1"
+	type: "Convolution"
+	bottom: "data"
+	top: "conv1"
+	param { lr_mult: 1.0 }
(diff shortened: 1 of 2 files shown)


utsavgarg:
@prakharchoudhary The prototxt does not load in Fabrik. Please test it on [fabrik.cloudcv.org](fabrik.cloudcv.org) before submitting a PR.

prakharchoudhary:
@utsavgarg I found the prototxt file for ZF net using faster rcnn on the following link:
https://github.com/rbgirshick/py-faster-rcnn/blob/master/models/pascal_voc/ZF/faster_rcnn_end2end/train.prototxt
Can you please have a look as to why this prototxt file is being termed invalid?

utsavgarg:
@prakharchoudhary That is because that prototxt is using a custom layer `SmoothL1Loss` which is not part of the standard Caffe package.

prakharchoudhary:
@utsavgarg `SmoothL1Loss` is necessary for the fast rcnn model. So should we wait until Caffe provides support for this layer? Or maybe we can use the fast-rcnn fork of caffe (here at: https://github.com/rbgirshick/caffe-fast-rcnn) in place of the standard version.

virajprabhu:
@utsavgarg Is SmoothL1Loss not a Python layer?
@prakharchoudhary In general, I don't think switching to third-party Caffe forks to support custom layers is a great idea in the long run. Also, please remember to resolve conflicts before you submit a PR.

utsavgarg:
@virajprabhu No it's not a **Python** layer but a custom layer implemented in `cpp/cuda`.

PR #114, opened 2017-09-26
Fix: Add CaffeNet model to FabrikZoo(#105)

prakharchoudhary:
@utsavgarg CaffeNet model tested and added.

### DIFF
--- a/example/caffe/CaffeNet.prototxt
+++ b/example/caffe/CaffeNet.prototxt
@@ -1,0 +1,400 @@
+name: "CaffeNet"
+layer {
+  name: "data"
+  type: "Data"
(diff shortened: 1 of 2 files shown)


utsavgarg:
@prakharchoudhary the model zoo already contains AlexNet which is fairly similar to CaffeNet. I would suggest adding models from different modalities like text, sound or different tasks such as captioning, vqa etc.

dexter1691:
@prakharchoudhary : Take a look at issue #115 . Those models will be very valuable to have in model zoo. I recommend starting from simple models for text like Char-RNN for which there are plenty of open source models out there for Keras etc.

Issue #115, opened 2017-09-26
Add RNN / LSTM examples for Keras in ModelZoo


dexter1691:
Keras supports recurrent layers like LSTM / RNN / GRU etc. Right now model zoo for Fabrik doesn't contain any examples of network that uses these layers. 

We should add a few examples to demonstrate the capability of using recurrent layers in a network and to test how import / export works for these layers.

Ram81:
@dexter1691 can I take this?

dexter1691:
@Ram81 : Go for it! Thanks!

prakharchoudhary:
@dexter1691 Since there are multiple models from several domains that can be added, would it be okay for me to work on this issue as well?

dexter1691:
@prakharchoudhary Ofcourse! Feel free to work on it. I think to ensure efforts are not duplicated, before working on adding the examples, mention the model here so that everyone is aware of it.

Ram81:
@dexter1691 can you point us to some of the models!

utsavgarg:
@Ram81 you could look at Captioning models (Show and Tell, Neuraltalk), VQA models, and Text generation models.

Ram81:
@utsavgarg I am referring this one https://github.com/jeffdonahue/caffe/blob/recurrent-rebase-cleanup/examples/coco_caption/lstm_language_model.prototxt as of now but I would need keras models right?

Ram81:
@utsavgarg would it be fine if I export this model https://github.com/fchollet/keras/blob/master/examples/lstm_text_generation.py to json then use it as an example?

utsavgarg:
@Ram81 you could do it in Caffe or Keras, but choose one out of these two as they are quite similar.

Ram81:
@utsavgarg I used this model https://github.com/jeffdonahue/caffe/blob/recurrent-rebase-cleanup/examples/coco_caption/lstm_language_model.prototxt  & tested it in here's the model

![lstm](https://user-images.githubusercontent.com/16323427/30950585-791ce8fe-a43b-11e7-8530-0e0029e4848b.png)
 But when I tried to export it to keras it says Can't determine size of HD5Data layer

Ram81:
Also there are some examples available but I would need to obtain a json file through its implementation if you want I can test them too.

utsavgarg:
@Ram81 When you are adding this network, can you also see how to handle the visualization for it better? Also, the error for export to Keras is expected so that is fine.

[...]

PR #120, opened 2017-09-29
Fix: Added CoCo caption model to demonstrate use of LSTM layer (#115)

Ram81:
@dexter1691 Fixed :+1:

### DIFF
--- a/example/caffe/coco_caption.prototxt
+++ b/example/caffe/coco_caption.prototxt
@@ -1,0 +1,150 @@
+name: "lstm_language_model"
+layer {
+  name: "data"
+  type: "HDF5Data"
+  top: "cont_sentence"
+  top: "input_sentence"
+  top: "target_sentence"
+  include { phase: TRAIN }
+  hdf5_data_param {
+    source: "./examples/coco_caption/h5_data/buffer_100/train_unaligned_batches/hdf5_chunk_list.txt"
+    batch_size: 20
+  }
+}
+layer {
+  name: "data"
+  type: "HDF5Data"
+  top: "cont_sentence"
+  top: "input_sentence"
+  top: "target_sentence"
+  include {
+    phase: TEST
+    stage: "test-on-train"
+  }
+  hdf5_data_param {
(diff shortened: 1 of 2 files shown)


Ram81:
model
Here's how model looks when loaded @utsavgarg

virajprabhu:
@Ram81 A bit part of Fabrik is being able to visualize models clearly. Can we try to have a better looking visualization?

Ram81:
@virajprabhu I tried changing it, can you suggest how do I change it so that it can be visualized better?

virajprabhu:
@utsavgarg You had generated good visualizations for some pretty complex models - can you guide @Ram81 ?

utsavgarg:
@Ram81 try thinking of it like this --> start from the deepest child node and try to make all of its parents fall at the same level. I am not sure it will work for all cases but should be alright. Let me know if you have troubles implementing it, I'll give it a try.

utsavgarg:
@virajprabhu @Ram81 I saw the visualization problem in more detail. I don't think there is a problem with the algo which decides the position, but the problem is coming when the top and bottom names of a layer are same and there are layers connected differently in different phases (train, test). After simplifying these issues in the prototxt the visualization seems to be working fine. Thought this might help you solve the problem.
image

Ram81:
@utsavgarg Thank you very much. Can you share the changes made or should I do it in my version on my own?

Ram81:
@utsavgarg Thank You very much. I understood I'll do the required changes & update the PR

Ram81:
@utsavgarg how should I handle separate model representation for both Train & Test, my Train model looks similar to your one
train

Ram81:
@utsavgarg @virajprabhu I updated the model please review
train
Train Model
test
Test Model

utsavgarg:
@Ram81 maybe I didn't explain properly, but I did not want you to change the `prototxt` to fit in the current visualization but rather the implementation to handle such `prototxt` files. In my last comment I was trying to explain why the visualization was failing, so please have a look at that.

Ram81:
@utsavgarg so should I open an issue for handling the visualization for such cases?

utsavgarg:
@ram81 closing this. Please submit another one after #127 is resolved.

PR #135, opened 2017-10-24
Fix: Handling Visualisation to avoid Overlapping Layers

Ram81:
@utsavgarg @deshraj please review
Existing Behaviour
abc
Updated Behaviour
abc
Above model is Inception V3


utsavgarg:
Nice work! :)

Issue #127, opened 2017-10-04
Handle visualization of models having different arrangement for Test, Train phases.


Ram81:
![test](https://user-images.githubusercontent.com/16323427/31188698-68e79aca-a953-11e7-95f1-cee87916db0e.png)
Test Model
![train](https://user-images.githubusercontent.com/16323427/31188714-7415302e-a953-11e7-89aa-e8a94cb52ce0.png)
Train Model

Ram81:
@utsavgarg I will start working on this :smiley:

Ram81:
@utsavgarg Any suggestions on this?

adityaXXX:
@utsavgarg Can I contribute by solving this issue?

Ram81:
@adityaXXX I am currently looking into various problems for this issue, can you please look for some other issue!

PR #138, opened 2017-10-31
Fix: Improve Check for Type of Phase

Ram81:
@utsavgarg as we don't consider stage arguments I thought a proper check for phase type is needed, plus it helped with the problem of separating phases in Visualisation issue #127 
Please review


Ram81:
@virajprabhu I updated this check to assign phase as we are not considering stage arguments, if you want the default phase to be train phase please let me know!

PR #146, opened 2017-11-10
Fix: Added YOLONet Object Detection Model

Ram81:
@utsavgarg Object Detection model added.
Reference: https://github.com/xingwangsfu/caffe-yolo/blob/master/prototxt/yolo_deploy.prototxt


Ram81:
@virajprabhu please review this PR too!

PR #149, opened 2017-11-13
Minor fix : favicon path

Jinsu-L:
image

Change path from "/images/favicon.ico" to "/static/img/favicon.ico".


Jinsu-L:
After clear commit,  resubmit

PR #151, opened 2017-11-14
FIX #148: Python ModelElement TypeError solve 

### LINKED ISSUE #148: Python ModelElement in topBar doesn't worked
Hello. I have some questions. When Python pane drags and drops, it doesn't work and occurs error. Maybe is it not implemented yet?

temp
image

GoodLuckDay:
```js
if (layer.info.type == 'Python'){
        // Changing endpoints depending on the type of Python layer
        if (layer.params.endPoint[0] == '1, 0'){
          data[layer.info.type]['endpoint']['trg'] = [];
          data[layer.info.type]['endpoint']['src'] = ['Bottom'];
        }
        else if (layer.params.endPoint[0] == '0, 1'){
          data[layer.info.type]['endpoint']['trg'] = ['Top'];
          data[layer.info.type]['endpoint']['src'] = [];
        }
        else{
          data[layer.info.type]['endpoint']['trg'] = ['Top'];
          data[layer.info.type]['endpoint']['src'] = ['Bottom'];
        }
        if (layer.params){
          Object.keys(layer.params).forEach(param => {
            if (param != 'endPoint'){
            data[layer.info.type]['params'][param] = {'name': param, 'type': 'text',
            'required': false, 'value': ''};
            }
          });
        }
      }
```
I handled this bug. The endpoint object does not exist in params, so It causes TypeError.
```js
layer.params['endPoint'] = [data[type]['endpoint'], false];
```
So I added this code in drop function.

### DIFF
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -52,15 +52,15 @@ class Canvas extends React.Component {
               and modify their border radius */
               if ($.inArray(net[outputId].info.type, combined_layers) != -1 &&net[inputId].connection.output.length==1){
                 if ($.inArray(net[inputId].info.type, combined_layers) == -1){
-                  $('#'+inputId).css('border-radius', '10px 10px 0px 0px') 
+                  $('#'+inputId).css('border-radius', '10px 10px 0px 0px')
                 }
                 else {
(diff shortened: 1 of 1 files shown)


utsavgarg:
LGTM :+1:

PR #161, opened 2017-12-04
Collapse sidebar on small screens

NSDCars5:
Google Code-In task. When screen width <= 1200px, sidebar cannot be accommodated on screen permanently. Here the sidebar is collapsed by default on such screens and a hamburger menu icon is present to show it.


NSDCars5:
Forgot to git add the content.js file in the previous commit, sorry about that.

virajprabhu:
@NSDCars5 LGTM, thanks for the PR!

PR #163, opened 2017-12-04
Make json_to_prototxt.py modular

owenl131:
Change Caffe implementation to be modular like Keras implementation


virajprabhu:
Thanks for the PR @owenl131 ! Since this is a large PR and part of an overall refactoring, could you submit this to the refactor/caffe branch instead?

PR #166, opened 2017-12-05
Added VQA

vpn1997:
Added VQA in model zoo.


ide/static/js/netLayout_vertical.js
```diff
@@ -33,6 +33,7 @@ function allocatePosition(layerId, preferredPosition){
   if (!map.hasOwnProperty(preferredPosition[1])) {
     map[preferredPosition[1]] = [];
   }
+  map[0]=[];
```
vpn1997: Added because without it,an error was showing like map[0](map[preferredPosition[1]-1) not defined.

ide/static/js/netLayout_vertical.js
```diff
@@ -33,6 +33,9 @@ function allocatePosition(layerId, preferredPosition){
   if (!map.hasOwnProperty(preferredPosition[1])) {
     map[preferredPosition[1]] = [];
   }
+    
+  map[0]=[];
```
vpn1997: map[0] was not define earlier(because preferredPosition[1] can never be 0) and because of that it was showing error in later code(map[preferredPosition[1]-1]).I dont really know why this error was not found earlier.May be because this model is unique and has 2 input sources.

package.json
```diff
@@ -41,6 +41,6 @@
     "file-loader": "^0.8.5",
     "style-loader": "^0.13.1",
     "url-loader": "^0.5.7",
-    "webpack": "^1.13.1"
+    "webpack": "^1.15.0"
```
vpn1997: I dont think it is strictly required .

ide/static/js/netLayout_vertical.js
virajprabhu: Good catch, thanks!

package.json
virajprabhu: Okay, can you roll it back? We shouldn't be unnecessarily updating dependencies.

vpn1997:
@virajprabhu Done.

virajprabhu:
Thanks @vpn1997. What VQA model is this? Can you share a link to the original?

vpn1997:
@virajprabhu https://github.com/iamaaditya/VQA_Demo

ide/static/css/style.css
```diff
@@ -355,6 +355,34 @@ body {
   padding: 0px;
 }
 
+.ReactModal__Content {
+  max-height: 90%;
+  width: 65%;
+  overflow: auto;
+}
+
```
vpn1997: These are the UI updated that was done in last pull request .

vpn1997:
@virajprabhu hey can you tell me why is it showing conflicting error?

utsavgarg:
@vpn1997 in the model zoo modal, please place VQA heading below that of Caption, not below Recognition. Also, the export for this model is not working. In the future, please ensure that the export to at least the framework from which the model was imported works.

keras_app/views/export_json.py
```diff
@@ -89,8 +89,8 @@ def export_json(request):
         dataLayers = ['ImageData', 'Data', 'HDF5Data', 'Input', 'WindowData',
                       'MemoryData', 'DummyData']
         processedLayer = {}
-        inputLayerId = None
-        outputLayerId = None
+        inputLayerId = []
```
vpn1997: So that it can handle more than one input source.

utsavgarg:
LGTM :+1:

PR #167, opened 2017-12-05
Make json_to_prototxt.py modular

owenl131:
Change Caffe implementation to be modular like Keras implementation


virajprabhu:
Thanks @owenl131, will review this as soon as possible.

PR #169, opened 2017-12-07
Update README - sudo -u -> sudo -i -u

trexdinosaur:
Proper usage from sudoing into another user should be -i . Just a -u  throws an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ This app is presently under active development and we welcome contributions. Ple
 5. [Install postgres >= 9.5](https://www.digitalocean.com/community/tutorials/how-to-install-and-use-postgresql-on-ubuntu-16-04)
 * Setup postgres database
     * Start postgresql by typing ```sudo service postgresql start```
-    * Now login as user postgres by running ```sudo -i -u postgres psql``` and type the commands below
+    * Now login as user postgres by running ```sudo -u postgres psql``` and type the commands below
     
     ```
       CREATE DATABASE fabrik;
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ This app is presently under active development and we welcome contributions. Ple
         Save, exit and then run
         ```
         source ~/.bash_profile
+        cd .. 
         ```
     * For Mac users
         * [Install Caffe](http://caffe.berkeleyvision.org/install_osx.html)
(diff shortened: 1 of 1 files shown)


PalashTanejaPro:
@trexdinosaur hey, have you tried replicating this issue though. I have installed 3 times with just -u and have faced no problems whatsoever.

trexdinosaur:
Which OS are you on?
________________________________

trexdinosaur:
Oh I see, you've chosen to directly open a psql with just that one command. That works, I thought the intent was to open another shell.

trexdinosaur:
However, if psql is already open, why do we have to run psql -c for each command? I'll try to redo it on my side, and I'll reopen the PR if need be. Thanks.

trexdinosaur:
ok I tested it some more. @PalashTanejaPro, your method works, but it maybe better to use normal SQL statements with a semicolon at the end. 

Two Reasons:
1) Running these SQL commands as a normal command, rather than invoking another psql shell, makes the commands more verbose and shows the errors. 
2) If someone is running on a postgresql server where postgres is not the user they are working with, they would have to change postgres with every single line of command, while if they were just pure commands, only one change would be needed to run it as a different user.

PalashTanejaPro:
@trexdinosaur I think that there is clearly some difference in the outputs that we are getting. Why don't you watch the [installation guide](https://www.youtube.com/watch?v=zPgoben9D1w) that I made, so that you can understand my perspective better?

trexdinosaur:
@PalashTanejaPro, I think I may be unclear what I am referring to, let me update the PR with what I mean. 
@5:37, you enter all those commands into an instance of psql, but because you are spawning another instance of it, within psql, there are no outputs. After each statement, there should be messages like ALTER table or failed, but because of the spawning, it's not happening.  I'll update the PR in a bit to what I am referring to.

trexdinosaur:
Updated!

JPJPJPOPOP:
@trexdinosaur I told PalashTanejaPro about this same issue, i.e. I needed to run sudo -i -u too. I really don't know why, as it never gave me an error, but if I don't, I don't get any outputs from psql.

The image below shows what I think should be the correct thing:

trexdinosaur:
@JPJPJPOPOP , you're doing something a bit different. @PalashTanejaPro originally wanted to spawn psql's within psql's. That was fine, just inefficient. @JPJPJPOPOP the reason you had to use -i is because you forgot to invoke a psql shell, which then requires you to use sudo -i to make to it interactive. If you click see file changed, I have added the method I think is best for the installation.

PalashTanejaPro:
@trexdinosaur Just to clear it up a little bit, it wasn't me who added those commands. They were added by the original creators of the project.

utsavgarg:
LGTM :+1:  Thanks, @trexdinosaur :).

PR #170, opened 2017-12-07
Error Message on 500 error from backend while exporting. 

trexdinosaur:
This code changes the behavior when the frontend gets a 500 error from the backend, it throws an error now, rather than just infinitely hanging on a loading screen.

I am getting a mixed tabs and spaces error though, what would be the easiest way to fix it? Are there automated tools to fix it?

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -231,9 +231,10 @@ class Content extends React.Component {
           }
           this.setState({ load: false });
         }.bind(this),
-        error() {
-          this.setState({ load: false });
-        }
+        error : function () {
+       	  this.setState({ load: false });
+	  this.addError("Error");
+        }.bind(this)
       });
     }
   }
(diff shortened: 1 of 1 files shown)


JPJPJPOPOP:
@trexdinosaur The reason you are getting mixed tabs and spaces is because you _are_ using both tabs and spaces in the same line, i.e. lines 235 and 236. It is easily fixed by just replacing the tabs with spaces.

PalashTanejaPro:
@trexdinosaur Since you asked for automated tools, If you are using sublime or atom there should be an option to convert tabs to spaces. You can try that!

utsavgarg:
@trexdinosaur is there any specific model which wasn't throwing an error message and was also not loading? Also, can you add something similar to import as well just to have uniformity.

utsavgarg:
LGTM :)

PR #174, opened 2017-12-10
On click, a new layer connects to bottom of node if possible.

nu11us:
If you have concerns, it works on my local version.

### DIFF
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -82,7 +82,7 @@ class Canvas extends React.Component {
       else{
         t = t[1];
       }
-      if (x2-x1==80) {
+      if (x2-x1==80) { //since only layers added through handleClick will be exactly 80px apart, we can connect those like this.
         instance.connect({
           source: s,
           target: t});
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -73,7 +73,7 @@ class Canvas extends React.Component {
     if(Object.keys(net).length>1){
       const x1 = parseInt(net[`l${this.props.nextLayerId-2}`].state.top.split('px'));
       const x2 = parseInt(net[`l${this.props.nextLayerId-1}`].state.top.split('px'));
-      if (x2-x1==100) {
+      if (x2-x1==80) {
         instance.connect({
           source: instance.getEndpoints(`l${this.props.nextLayerId-2}`)[0],
           target: instance.getEndpoints(`l${this.props.nextLayerId-1}`)[1]});
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -35,8 +35,8 @@ class Canvas extends React.Component {
         grid: [8, 8]
       }
     );
+    const net = this.props.net;    
     if (this.props.rebuildNet) {
-      const net = this.props.net;
       let combined_layers = ['ReLU', 'LRN', 'TanH', 'BatchNorm', 'Dropout', 'Scale'];
       Object.keys(net).forEach(inputId => {
         const layer = net[inputId];
@@ -69,6 +69,15 @@ class Canvas extends React.Component {
       });
       this.props.changeNetStatus(false);
       // instance.repaintEverything();
+    }
+    if(Object.keys(net).length>1){
+      const x1 = parseInt(net[`l${this.props.nextLayerId-2}`].state.top.split('px'));
+      const x2 = parseInt(net[`l${this.props.nextLayerId-1}`].state.top.split('px'));
+      if (x2-x1==100) {
+        instance.connect({
+          source: instance.getEndpoints(`l${this.props.nextLayerId-2}`)[0],
+          target: instance.getEndpoints(`l${this.props.nextLayerId-1}`)[1]});
+      }
     }
   }
   allowDrop(event) {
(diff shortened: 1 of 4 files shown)


ide/static/js/canvas.js
```diff
@@ -70,6 +70,15 @@ class Canvas extends React.Component {
       this.props.changeNetStatus(false);
       // instance.repaintEverything();
     }
+    if(Object.keys(net).length>1){
+      const x1 = parseInt(net[`l${this.props.nextLayerId-2}`].state.top.split('px'));
+      const x2 = parseInt(net[`l${this.props.nextLayerId-1}`].state.top.split('px'));
+      if (x2-x1==100) {
```
utsavgarg: change `100` to `80`

ide/static/js/content.js
```diff
@@ -66,6 +67,8 @@ class Content extends React.Component {
     this.zooModal = this.zooModal.bind(this);
     this.modalContent = null;
     this.modalHeader = null;
+    this.clicker = false;
```
utsavgarg: Where is this being used?

ide/static/js/content.js
```diff
@@ -680,6 +683,47 @@ class Content extends React.Component {
     this.modalContent = <ModelZoo importNet={this.importNet}/>;
     this.openModal();
   }
+  
+  handleClick(event) {
+    this.placeholder = false;
```
utsavgarg: Remove this

ide/static/js/content.js
```diff
@@ -680,6 +683,47 @@ class Content extends React.Component {
...
+    event.preventDefault();
+
+    const net = this.state.net;
+    const id = event.target.id;
+    const prev = net[`l${this.state.nextLayerId-1}`];
+    const next = data[id];
+    const zoom = instance.getZoom();    
+    const layer = {};
+    let phase = this.state.selectedPhase;
+
```
utsavgarg: Need to add a case to handle situation when there is no layer, can do something like this
```
// To handle case when no layer has been added
    if (Object.keys(net).length==0){
      layer.connection = { input: [], output: [] };
      layer.info = {
            type: id.toString(),
            phase,
            class: ''
          }
      layer.params = {'endPoint': [next['endpoint'], false]};        
      Object.keys(next.params).forEach(j => {
        layer.params[j] = [next.params[j].value, false];
      });    
      layer.props = JSON.parse(JSON.stringify(next.props))
      const height = Math.round(0.05*window.innerHeight, 0);
      const width = Math.round(0.35*window.innerWidth, 0);
      var top = height + Math.ceil(41-height);
      var left = width;
      layer.state = {
            top: `${top}px`,
            left: `${left}px`,
            class: '' 
          }
      layer.props.name = `${next.name}${this.state.nextLayerId}`;          
      this.addNewLayer(layer);
      this.clicker = true;
    }
```

ide/static/js/content.js
```diff
@@ -680,6 +683,47 @@ class Content extends React.Component {
...
+    const net = this.state.net;
+    const id = event.target.id;
+    const prev = net[`l${this.state.nextLayerId-1}`];
+    const next = data[id];
+    const zoom = instance.getZoom();    
+    const layer = {};
+    let phase = this.state.selectedPhase;
+    
+    if (this.state.nextLayerId>0) {
```
utsavgarg: need to make this `else if`

ide/static/js/content.js
```diff
@@ -680,6 +683,47 @@ class Content extends React.Component {
...
+      if (prev.params.endPoint[0].src == "Bottom"){
```
utsavgarg: change to `if (data[prev.info.type].endpoint.src == "Bottom"){`

ide/static/js/content.js
```diff
@@ -680,6 +683,47 @@ class Content extends React.Component {
...
+        if (next.endpoint.trg == "Top") {
+
+          layer.connection = { input: [], output: [] };
+          layer.info = {
+            type: id.toString(),
+            phase,
+            class: ''
+          }
+          layer.params = {}
+          layer.params['endPoint'] = [next['endpoint'], false];
```
utsavgarg: remove this

ide/static/js/content.js
```diff
@@ -680,6 +683,47 @@ class Content extends React.Component {
...
+            class: ''
+          }
+          layer.params = {}
+          layer.params['endPoint'] = [next['endpoint'], false];          
+          Object.keys(next.params).forEach(j => {
+            layer.params[j] = [next.params[j].value, false];
+          });    
+          layer.props = JSON.parse(JSON.stringify(next.props))
+          layer.state = {
+            top: `${(parseInt(prev.state.top.split('px')[0])/zoom + 100)}px`,
```
utsavgarg: change `100` to `80`

ide/static/js/content.js
```diff
@@ -728,6 +774,7 @@ class Content extends React.Component {
             error={this.state.error}
             dismissError={this.dismissError}
             addError={this.addError}
+            clicker={this.clicker}
```
utsavgarg: I see this being passed to `canvas` but don't see it being used there.

ide/static/js/content.js
nu11us: Leftover that I forgot to delete.

utsavgarg: Still haven't deleted this?

PR #176, opened 2017-12-12
Ui changes - allows user to click on the row, and not just the arrow

trexdinosaur:
Ui changes - allows user to click on the row, and not just the arrow

Also, can I change up how the pane works, and have the data be served from django and then rendered in a React component? I think it is a bit redundant right now. 

Thanks.


JPJPJPOPOP:
@trexdinosaur Could you also try fixing a bug with the arrows? Currently if you select on an arrow and activate a dropdown, but then select another row's dropdown, the dropdown reverts, but the arrow still points down.

screen shot 2017-12-11 at 11 51 59 pm

utsavgarg:
Also, the colour of the menu items should not change.
image

And any changes to make the code more efficient are welcome, so go ahead :+1:

trexdinosaur:
PR has been updated @utsavgarg!

utsavgarg:
@trexdinosaur when you click in between the name and the arrow, the arrow symbol changes but the menu is not expanded. If we are changing it, we should change it so that a click anywhere on the row should expand the menu.

image

trexdinosaur:
@utsavgarg updated!

Issue #126, opened 2017-10-02
Display parameters of the model being visualized


utsavgarg:
We should display the number of parameters(weights of the model) the model contains somewhere on the screen. These parameter values should change whenever a new layer is added or some existing layer is changed.

gpowale:
You can assign this to me, I can get started on this issue :)

utsavgarg:
@gpowale sure go ahead.

ayush1999:
@utsavgarg @RishabhJain2018  Can I take up this issue? Please help me in getting started.

deshraj:
@ayush1999 please go ahead. We haven't heard from @gpowale for last 12 days.

ayush1999:
Thanks @deskraj . Can you please tell me how can I solve this? Just a little background.

utsavgarg:
@ayush1999 you will have to calculate the number of parameters when a model is imported which should be simple as you can calculate the size of the weight and bias matrices. And then whenever a layer is added or deleted you add or subtract its parameters from the total count.

Ram81:
@utsavgarg can I take this?

utsavgarg:
@Ram81 I'd recommend working on one issue at a time so that it's not blocked for others, so please finish #127 first as you had expressed interest in it and if still there is no update from @ayush1999 you can work on this.

Ram81:
@utsavgarg Sure

Ram81:
@utsavgarg can I start with this one now?

utsavgarg:
@Ram81 Sure.

kurianbenoy:
Is this issue fixed or can someone take it up?

Ram81:
@kurianbenoy I am currently working on it.

PR #179, opened 2017-12-13
Fix: Show Total Parameters of Model Visualized(#126)

Ram81:
@utsavgarg its the initial work for #126.
abc

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -192,8 +192,8 @@ class Content extends React.Component {
     });
     this.setState({ net, selectedLayer: null });
   }
-    updateParameters(layer) {
-    // obtain the total parameters of the model computed till now
+  updateParameters(layer) {
+    // obtain the total parameters of the model
     var totalParameters = this.state.totalParameters;
     var weight_params = 0;
     var bias_params = 0;
@@ -201,6 +201,7 @@ class Content extends React.Component {
     var filter_layers = ["Convolution", "Deconvolution"];
     var fc_layers = ["InnerProduct", "Embed", "Recurrent", "LSTM"];
 
+    // Currently we don't handle layers in which we don't need to consider bias
     if(filter_layers.includes(layer.info.type)) {
       // if layer is Conv or DeConv calculating total parameter of the layer using:
       // N_Input * K_H * K_W * N_Output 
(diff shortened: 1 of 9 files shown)


ide/views.py
```diff
@@ -1,5 +1,33 @@
...
+
+@csrf_exempt
+def calculate_parameter(request):
+    if request.method == 'POST':
+        net = yaml.safe_load(request.POST.get('net'))
+        try:
+            netObj = copy.deepcopy(net)
+            for layerId in netObj:
+                for param in netObj[layerId]['params']:
+                    netObj[layerId]['params'][param] = netObj[layerId]['params'][param][0]
```
utsavgarg: Add a comment explaining why is this required

ide/views.py
```diff
@@ -1,5 +1,33 @@
+import sys
```
utsavgarg: Please arrange them in alphabetical order

ide/utils/shapes.py
```diff
@@ -110,6 +110,50 @@ def repeat(layer):
     return shape
 
 
+def handle_concat_layer(outputId, inputId, net):
```
utsavgarg: Can't we have a function similar to other layers named `concat` and handle it inside `get_layer_shape`? Do we really need to handle it separately?

ide/static/js/content.js
```diff
@@ -293,6 +369,7 @@ class Content extends React.Component {
       success: function (response) {
         if (response.result === 'success'){
           this.initialiseImportedNet(response.net,response.net_name);
+          this.loadLayerShapes();
```
utsavgarg: Also, if you are calculating the shapes during import, can you remove the additional calls made to `get_shapes` during export in `export_json`.

Ram81: Yeah Sure

ide/utils/shapes.py
Ram81: This is the case for adding the num_output of all the input layers of concat layer, it is needed in topological sort part of shapes.py, where it was being used two times that's why to avoid redundancy I made a separate function

utsavgarg: Okay.

utsavgarg:
This calculates the total number of parameters after a model has been imported but does not calculate the change in those parameters after a layer is added or deleted. Parameter calculation has been tested for AlexNet, VGG-16, ResNet-101 and Inception-v3.

utsavgarg:
LGTM, Nice Work :)

Issue #154, opened 2017-11-20
 When exporting multiple times error


kdongha:
I found a bug.

-Step to reproduce the problem

1. The model is loaded.

 2. Export the model once.

 3. Export the model again.

Then. An error occurs.

The error message is:
bundle.js:44086 Uncaught TypeError: Cannot read property '0' of undefined
    at bundle.js:44086
    at Array.forEach (<anonymous>)
    at Tooltip.render (bundle.js:44081)
    at bundle.js:16306
    at measureLifeCyclePerf (bundle.js:15586)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (bundle.js:16305)
    at ReactCompositeComponentWrapper._renderValidatedComponent (bundle.js:16332)
    at ReactCompositeComponentWrapper._updateRenderedComponent (bundle.js:16256)
    at ReactCompositeComponentWrapper._performComponentUpdate (bundle.js:16234)
    at ReactCompositeComponentWrapper.updateComponent (bundle.js:16155)

virajprabhu:
Thanks for reporting, @kdongha!

vpn1997:
@virajprabhu I have sent a PR regarding this issue.

utsavgarg:
Fixed in #181

PR #181, opened 2017-12-16
Fix: Multiple Model Export(#154)

Ram81:
@utsavgarg #154 Fixed


Ram81:
@utsavgarg For fixing #154 I modified update of the net object's param attribute & deletion of state attribute by moving it to backend so that net state remains the same after export & we retrieve all our parameters, for that we needed to update shapes.py  for param attribute access.

utsavgarg:
LGTM, thanks !

PR #183, opened 2017-12-16
README Error

nu11us:
You cannot create a git repo when there is an existing folder of the same name. Switching the two steps removes this problem.


utsavgarg:
This problem only exists if you clone Fabrik to your home folder, a better fix would be to rename the `virtualenv` directory to be `fabrik`.

virajprabhu:
@nu11us Are you still working on this?

nu11us:
;

trexdinosaur:
@nu11us, the issue still exists for anyone cloning in their home directory... I think you should implement @utsavgarg's suggestion as that's a simple fix, and it will avoid any conflicting file issues.

PR #185, opened 2017-12-17
Fix #117: Add a title pane

### LINKED ISSUE #117: Add a title pane
We should add a title pane with the model name – this should be an editable field which:
* defaults to untitled if the user is building a model from scratch
* takes the name of whatever the model file is called if loading from zoo / importing from any framework
* Should go into DB – when a shared link is opened, the same name should show up.

NSDCars5:
The default name for a net is Untitled. When a net is loaded from the model zoo, its name is loaded into the textbox provided, which can be changed, and the changed name will be reflected in the exported file.

### DIFF
--- a/ide/static/css/style.css
+++ b/ide/static/css/style.css
@@ -88,9 +88,19 @@ body {
 input#netName {
   width: 100%;
   border: none;
-  padding: 5px 10px;
-  border-radius: 5px;
-  background: white;
+  position: absolute;
+  top: 1.5rem;
+  left: 1rem;
+  font-size: 2rem;
+  padding: 0 10px;
+  outline: none;
+  z-index: 100;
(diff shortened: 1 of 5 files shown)


utsavgarg:
Nice work :+1: Can you change the position of the net name field to be on the canvas on top of the network.

NSDCars5:
@utsavgarg Made the required changes.

utsavgarg:
@NSDCars5 please centre it on the canvas and add a faint underline so it's clearer that it's editable

NSDCars5:
Not sure why the Travis build is failing, I only changed the UI...

utsavgarg:
Please merge with master, there was a small bug which was recently fixed which might have caused the build to fail.

PR #186, opened 2017-12-18
Visualization algo - Alpha changes 

trexdinosaur:
Still needs testing!


PalashTanejaPro:
@trexdinosaur Would love to test this algorithm, can you add comments explaining it though?

trexdinosaur:
@PalashTanejaPro Yep, that's on my todo list! I just want to make sure the algorithm is working well and then I will comment the code code and clean it up, it is very messy currently.

trexdinosaur:
So I've already found a major bug: I'm relying on the layer id to give order of layers, but it always doesn't work out this way, so i will need to recode a little bit of the algo.

utsavgarg:
@trexdinosaur even this doesn't seem to be working for the two failing cases in the last algorithm i.e. MNIST Siamese and DenseNet (refer to the previous PR for details). It also seems to break a few cases that were working before (below is seq2seq)
image.
You are not up to date with the master branch and some changes are breaking the drag and drop functionality of layers(unable to connect any two dragged layers).

trexdinosaur:
@utsavgarg, I have made more changes :)

It should work with the models you have mentioned, and I have made the algorithm much smarter at how it handles the entire process. However, this has made it very slow, and on the dense net, I can barely move around elements. I will make it more efficient soon to only recalculate every 1 second or 2 seconds when dragging around nets that are very large. Can you please test and let me know if there needs to be changes to any specific nets?

Thanks.

PalashTanejaPro:
@trexdinosaur Is this algorithm correct 100% of the time? Because Fabrik is not only intended to be used with just the zoo models. 
Even though I agree that any improvement is a step in the right direction, but I don't think that a mathematical algorithm will suffice and will need complete replacement further down the road.

trexdinosaur:
@PalashTanejaPro 

What do you suggest if a mathematical algorithm is not a best option? Even the more complex algorithms that are used for tree-based displays are just using math behind their reasoning? My algorithm is more rudimentary than more famous algos that are good at positioning, but it does handle the following aspects:

- It tries to detect if a line is cutting through a node and if it is:
- it will iteratively move over in the direction that is less crowded, and if it is even, it (should) default to right.


Are you suggesting implementing a better algorithm that is not math based? I haven't heard of that type of algorithm, so please let me know and I can go ahead and implement it. 

Otherwise, if I can improve my algo a bit, please let me know and I would happy to implement those changes :) It is not perfect 100% time, but it is a definite improvement from before!

PalashTanejaPro:
@trexdinosaur First off, there definitely are more complex algorithms that are not mathematical; that is if you do not consider programming to be a subset of mathematics. Would suggest looking up graph algorithms if you have the time to do so.

Also, the main point I was trying to make was that sometimes a new algorithm that we make is actually better only in some cases and worse in others. 

All I was saying was that do you have an understanding of how this algorithm behaves on edge cases. Because if the algorithm is not provable to be 100% correct then it needs very extensive testing to understand its behaviour on edge cases.

trexdinosaur:
@PalashTanejaPro I see, there are definitely a few cases that are causing issues, and I am trying to pin and fix them. 

Maybe we can try finding a good algorithm that works well for this scenario. Do you want to discuss more over the Gitter?

utsavgarg:
@trexdinosaur Great work overall :+1: A few points for improvement:
* Add some comments explaining the functions you have added, it's easier for others who try to build on this in the future. 
* This is **breaking the drag and drop feature**, I am unable to connect layers manually, we definitely need to fix that. 
* Reduce the horizontal displacement of arrows/connectors
* Please fix the merge conflicts.

trexdinosaur:
> Add some comments explaining the functions you have added, it's easier for others who try to build on this in the future.
This is breaking the drag and drop feature, I am unable to connect layers manually, we definitely need to fix that.
Reduce the horizontal displacement of arrows/connectors
Please fix the merge conflicts.

1) Done
2) Done
3) Done
4) Done
5) Done 

:) I'll probably write some documentation rather than just comments, as I think it will be easier to follow, and I'll have some more space.

ide/static/js/layoutCheckAlgo.js
```diff
@@ -0,0 +1 @@
+asdf
```
virajprabhu: Do we need this file?

virajprabhu:
This looks mostly good to me. Drag and drop seems to breaking sometimes, but I'm not sure if this PR introduced that. For eg. I'm unable to make connections with Python layers.

trexdinosaur:
@virajprabhu it was caused by my PR, it is updated now! file has been deleted

utsavgarg:
LGTM :+1:

PR #187, opened 2017-12-18
Add unit tests for caffe DB and keras import_json

owenl131:


utsavgarg:
@owenl131 thanks for the PR, will review soon.

owenl131:
Hi @utsavgarg how is it?

tests/unit/caffe_app/test_db.py
```diff
@@ -0,0 +1,52 @@
...
+        tests = open(os.path.join(settings.BASE_DIR, 'tests', 'unit', 'ide',
+                                  'caffe_export_test.json'), 'r')
+        net = json.load(tests)['net']
+        response = self.client.post(
+            reverse('saveDB'),
+            {'net': net, 'net_name': 'netname'})
+        response = json.loads(response.content)
+        self.assertEqual(response['result'], 'success')
+
+    def test_save_jsonstring(self):
```
utsavgarg: This is not required.

tests/unit/keras_app/test_views.py
```diff
@@ -35,17 +38,40 @@ def setUp(self):
...
+    def test_keras_import_sample_id(self):
+        response = self.client.post(
+            reverse('keras-import'),
+            {'sample_id': 'vgg16'})
+        response = json.loads(response.content)
+        self.assertEqual(response['result'], 'success')
+        self.assertEqual(response['net_name'], 'vgg16')
+        self.assertTrue('net' in response)
+
+    def test_keras_import_invalid_sample_id(self):
```
utsavgarg: Either split the Caffe `sample_id` tests into two or combine these into one for uniformity

tests/unit/caffe_app/test_db.py
owenl131: Which part should be removed?

utsavgarg: The `test_save_jsonstring` function as both of them cover the same code.

owenl131: One is with the net passed in as a string while the other is passed in as an object. Not sure whether there is any difference in the underlying mechanism.

owenl131: Updated :D

utsavgarg:
LGTM :)

PR #189, opened 2017-12-18
Fix #188: Collapsibles' arrows

### LINKED ISSUE #188: Collapsibles' arrow in the sidebar can get unsynchronized
When you press an arrow twice, or you have expanded one category in the sidebar and expand another category, the arrows will get unsynchronized.

adamstafa:
`toggleClass()` now iterates over all categories and sets the real state from `aria-expanded` attribute to each var in this.state.


utsavgarg:
LGTM :)

PR #190, opened 2017-12-18
Fix #184: Export fails after layer is added

### LINKED ISSUE #184: Export fails after layer is added
Export fails if a new layer is added to a model that has been loaded from the Fabrik model zoo.
Steps to recreate:
-  Load YoloNet from the Fabrik model zoo
- Add a Softmax loss layer
- Click on export to Caffe

NSDCars5:
The endPoint property seems to be causing problems, as this information is Fabrik-specific and not related to Caffe.

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -199,7 +199,7 @@ class Content extends React.Component {
       Object.keys(layer.params).forEach(param => {
         layer.params[param] = layer.params[param][0];
         const paramData = data[layer.info.type].params[param];
-        if (layer.info.type == 'Python' && param == 'endPoint'){
+        if (layer.info.type == 'Python' || param == 'endPoint'){
           return;
         }
         if (paramData.required === true && layer.params[param] === '') {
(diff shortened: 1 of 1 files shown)


utsavgarg:
@NSDCars5 This doesn't seem to fix the bug, I am still getting the same error. Were you able to reproduce the error by following the steps mentioned in the issue?

NSDCars5:
@utsavgarg Yes, I was, and making this change fixed it on my end - I was able to export a prototxt and re-import it into Fabrik.

utsavgarg:
@NSDCars5 After this change, it is indeed exporting a model, but the exported model does not contain the added layers.

NSDCars5:
@utsavgarg This is what I get: https://www.youtube.com/watch?v=Dbv4is0EFGM

Does the position of the layers have to be preserved? Other than that it seems to work fine for me, the new layers are shown once I import it.

utsavgarg:
@NSDCars5 It works if you drag and drop a layer but doesn't work if you click and add a layer. You can see the video below to get a better understanding of what I mean
https://drive.google.com/open?id=124G_gKPe0CO3K4T9X3wcBkjkbQTqAv3j

NSDCars5:
@utsavgarg Hey, I think this is a problem with the handleClick function not properly joining a layer added manually. Working on a fix.

NSDCars5:
@utsavgarg Should work fine now.

utsavgarg:
LGTM :)

PR #192, opened 2017-12-18
Fix: title no longer overlaps with error messages, if any.

NSDCars5:
Previous:

errorbelowname

Current:

errorabovename


utsavgarg:
Hi, there is another small issue, when you create a network by clicking to add layers, the title overlaps with the first layer (we should change the top offset for the first layer)

image

NSDCars5:
@utsavgarg Made the above change, now there is no overlap.

utsavgarg:
LGTM :)

PR #196, opened 2017-12-18
Fix #155: Parameters tooltip not displayed

### LINKED ISSUE #155: The layer parameter doesn't displayed.
When I drag and drop any layer, the parameter is shown well.
But if new layer is dragged and dropped, that it doesn't display parameter.
Except for the first layer, the other layers don't show the parameters.

----------------------------------------------------------------------------

adamstafa:


utsavgarg:
LGTM :)

PR #202, opened 2017-12-19
Fix #199: Cannot add layer after delete

### LINKED ISSUE #199: Cannot add a layer after delete
A layer cannot be added by click/drag and drop after a layer has been deleted.

### Steps to reproduce
* Add an input layer
* Add any layer and then delete it
* Try to add any other layer

### Expected behaviour
After a layer has been deleted, any new layer the user wishes to add should connect to the previous layer

### Observed behaviour
After a layer has been deleted, the user is unable to add any other layer

adamstafa:
There was a problem `deleteLayer()` in `content.js`. `nextLayerId` stayed the same even after the last layer was deleted. My fix substracts 1 from nextLayerId if last layer was deleted.

~~There are still some issues when layer in middle is deleted, you can't delete layers in the separated part of the net anymore. You can still add layers to the end tho.~~

edit: It's actually a problem of indexing. Some methods expected that the layers will be in a sequence (no index missing). Deleting a layer caused missing index which caused the bug.

### DIFF
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -73,10 +73,24 @@ class Canvas extends React.Component {
     }
     // Might need to improve the logic of clickEvent
     if(Object.keys(net).length>1 && this.props.clickEvent){
-      const x1 = parseInt(net[`l${this.props.nextLayerId-2}`].state.top.split('px'));
-      const x2 = parseInt(net[`l${this.props.nextLayerId-1}`].state.top.split('px')); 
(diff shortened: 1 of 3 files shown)


adamstafa:
I modified the algorithm for connecting layers added from dropdown by clicking. It now searches for 2 non-undefined layers from bottom of the net insted of getting 2 last ids which can be already deleted. Still it would be nice to be able to reindex layers.

utsavgarg:
LGTM :+1:

PR #204, opened 2017-12-19
Fix #203: Layers added in correct position regardless of scroll

### LINKED ISSUE #203: Layers added in wrong position after scroll
**Steps to reproduce:** 

1. Add a few layers, distance them such that you have to scroll to see all of them.

2. Scroll down, and add a layer by dragging and dropping.

**Expected outcome:** Layer is added where you dropped it.

**Observed outcome:** Layer is added somewhere near the top of the canvas, where you have to scroll up to see it.

I think I have a fix, I'll start working on it.

NSDCars5:
On drag and drop, layers used to be added on the basis of the position of the mouse cursor relative to the top of the viewport. This would cause a problem when the canvas scrolled, as the viewport did not scroll with it. Scrolling offsets have been added so this problem doesn't occur anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ This app is presently under active development and we welcome contributions. Ple
 6. Install node modules
 ```
 npm install
+sudo npm install -g webpack
 webpack --progress --watch --colors
 ```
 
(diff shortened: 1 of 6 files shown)


utsavgarg:
LGTM :+1:

PR #205, opened 2017-12-21
Add Inception-v2 to model zoo.

nu11us:
I added an ImageNet model to the model zoo. 
I found it on caffe's Model Zoo [here](https://github.com/BVLC/caffe/wiki/Model-Zoo#inception-bn-full-imagenet-model). It imported fine both on my own host and at [fabrik.cloudcv.org](fabrik.cloudcv.org)

### DIFF
--- a/example/inception.prototxt
+++ b/example/inception.prototxt
@@ -1,0 +1,3065 @@
+name: "Inception21k"
+layer { 
+  name: "data"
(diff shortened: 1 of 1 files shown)


nu11us:
I didn't see that we had Inception V2. Sorry about that.

Issue #206, opened 2017-12-22
Cyclic edges should not be created


Ram81:
@utsavgarg Even though Fabrik detects cyclic edges but it still doesn't delete them. In case we detect a cyclic edge we should not allow it to be created.
![abc](https://user-images.githubusercontent.com/16323427/34289978-1392bf68-e71b-11e7-823d-9c9ccee2f6c3.png)

Ram81:
@utsavgarg even when we add layers using click event the logic is not able to detect cyclic edges, because we aren't adding parent layers of the newly added layers.

utsavgarg:
@Ram81 Detecting cyclic edges while creating the graph will cause inefficiencies and might not be ideal, the current logic detects them before export which I think is alright. Also, is there a sequence of actions that can lead to the creation of cyclic edges using click events?

Ram81:
No, there is no sequence of actions that lead to the cyclic edges, the same problem due to which export was failing the new layer added using click event wasn't being stored in net object that's why for a layer added using click event cycle was not getting detected.

Ram81:
Also I think I didn't specify this properly, we should detect cycles while creating graph, I meant if we detect a cycle then we should delete the connection added by the cyclic edge, as you can see in the snapshot the edge from pool layer to conv layer is still present even if it is detected as cycle. We should remove the new edge that is causing cycles in model (even from frontend).

PR #207, opened 2017-12-22
Fix: Handle Addition of Layer using Click Event(#206)

Ram81:
Each time we were adding a new layer to existing model using click event input & output layer connection was being drawn using frontend but net object did not contain a connection between layers.


utsavgarg:
@Ram81 I think this was fixed in #190

PR #208, opened 2017-12-22
Added HED

nu11us:


utsavgarg:
@nu11us You haven't added the model to the front-end. Also, the export to Caffe for this model is throwing an error. In the future, please check if export to at least the framework used for import works fine.

utsavgarg:
LGTM :)

PR #209, opened 2017-12-23
GCI: Fix #119: Add an image captioning model

### LINKED ISSUE #119: Add Image Captioning examples in model zoo
Image Captioning is another popular task for which we should have an example in Model Zoo

galiaxy:
I made this for #119.

The image captioning model was from https://github.com/LemonATsu/Keras-Image-Caption .


utsavgarg:
Not merging, as the model is not as popular.

PR #210, opened 2017-12-23
Fix #198: Fabrik now works on Edge

### LINKED ISSUE #198: Fabrik Doesn't work on Edge
Fabrik doesn't seem to be working on the latest version of edge. Seems to be an issue w/ jsplumb.

As a temporary fix, can I add a message that says please use Google Chrome to access Fabrik?

NSDCars5:
**Issue 1: Dragging/dropping layers did not work:**

This issue was caused due to the spec set for setData/getData in drag-drop events, where the data is stored in a "protected mode", ie. no event other than the original one is allowed to access it. Firefox and Chrome don't follow the W3C spec so setData/getData works there, but Edge follows the spec and doesn't allow it.

More info here: https://stackoverflow.com/questions/11927309/html5-dnd-datatransfer-setdata-or-getdata-not-working-in-every-browser-except-fi

The most effective solution is to just use a different means of storing what is being dragged or dropped, so I used a React state variable.

**Issue 2: Models could not be imported from file:**

This seems to have been caused due to the fact that while every other browser sends a .prototxt file with content-type application/octet-stream, Edge sends it with text/plain. Simply allowing text/plain files to be imported as models works, and invalid files will still be caught by the try/except blocks.

### DIFF
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -544,7 +544,8 @@ layer_dict = {'Accuracy': Accuracy,
 def import_prototxt(request):
     if request.method == 'POST':
         if ('file' in request.FILES) and \
-           (request.FILES['file'].content_type == 'application/octet-stream'):
+           (request.FILES['file'].content_type == 'application/octet-stream' or \
+            request.FILES['file'].content_type == 'text/plain'):
             try:
                 prototxt = request.FILES['file']
             except Exception:
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -544,8 +544,8 @@ layer_dict = {'Accuracy': Accuracy,
 def import_prototxt(request):
     if request.method == 'POST':
         if ('file' in request.FILES) and \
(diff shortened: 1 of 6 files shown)


utsavgarg:
@NSDCars5 We faced some issues with exports, where the `number` text fields are being treated as plain text fields in Edge, and it causes data type issues in the backend. Can you see if there is a way to fix this?

NSDCars5:
@utsavgarg I'll check. Can you point me to an example?

utsavgarg:
@NSDCars5 For example, the data layer in COCO Caption has a batch size argument.

NSDCars5:
@utsavgarg Is this where changing the batch size results in an "Number has type str, but expected one of: int, long"? I can reproduce this even in Chrome, this may not be a bug in Edge in particular. 

I fixed the problem, but since it's not related to Edge, should I make it a separate PR?

virajprabhu:
@NSDCars5 A separate PR would be great! After we merge that, can you pull upstream changes and  we'll try to see if there are still outstanding issues with Windows?

NSDCars5:
@vijayprabhu No problem, will do.

utsavgarg:
LGTM :)

PR #211, opened 2017-12-23
Added DRN VQA

nu11us:
This is Deep Residual Networks VQA from [here](https://github.com/KaimingHe/deep-residual-networks). 

It's pretty big, so it takes time to load, but I'm pretty sure it works.


nu11us:
Please ignore this.

PR #212, opened 2017-12-25
Added FCN for Semantic Segmentation

nu11us:


utsavgarg:
@nu11us Any captioning model is not as useful without the decoder portion(RNN or LSTM) of the network, but the model you have added just contains the CNN(encoder) without the Recurrent portion. So, either complete the model or look at a different one. Also, please add the source from where you get the model.

nu11us:
The RNN for Neuraltalk does not use caffe, so rather than recreating it by hand, I've chosen to replace it with this [one](https://github.com/shelhamer/fcn.berkeleyvision.org) from the main branch of caffe's model zoo. While it does not include any recurrent layers, it does include a different ending like: 
eltwise
This model seems closer to Scene Labeling than Image Captioning, but since both take image inputs and give text descriptions, I think it is fine. Both imports and exports through caffe worked when tested.

utsavgarg:
@nu11us We can accept this. But, this model is for semantic segmentation, not image captioning which is a different task. So, change the title for this PR and put the model under a different heading in the Model Zoo modal. Also, the UI of the modal is getting disturbed by this, just check that it works fine for different resolution screens.

nu11us:
Switched current task and model category.

utsavgarg:
image
Change the model zoo UI back to have 5 columns so that there need to scroll.

nu11us:
Okay.

utsavgarg:
LGTM :)

PR #213, opened 2017-12-26
Change scroll on header

trexdinosaur:
changes on how scrolling side bar works, doesn't smush against the top part anymore


utsavgarg:
@trexdinosaur can you add some more details on this, explaining what the issue was using screenshots etc.

PR #219, opened 2017-12-28
Added Dex-Gender Model

gautamjajoo:
Added a new model DEX-Gender Estimation Model (Deep EXpectation of Gender) . Made the changes in front-end also. 
@utsavgarg

### DIFF
--- a/example/caffe/gender.prototxt
+++ b/example/caffe/gender.prototxt
@@ -1,0 +1,595 @@
+name: "DEX-GENDER"
+layer {
+  top: "data"
+  type: "ImageData"
+  top: "label"
+  name: "data"
+  transform_param {
+    mirror: true
+    crop_size: 224
+    mean_file: "imagenet_mean.binaryproto"
+  }
+  image_data_param {
+    source: "train.txt"
+    batch_size: 10
+    new_height: 256
+    new_width: 256 
+  }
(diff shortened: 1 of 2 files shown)


virajprabhu:
@gautamjajoo Thanks! Can you add a link to the original prototxt / repo?

gautamjajoo:
@virajprabhu https://data.vision.ee.ethz.ch/cvl/rrothe/imdb-wiki/

utsavgarg:
@gautamjajoo Nice effort. A few suggestions:
* **Export is failing**, please always check before adding models to the Fabrik model zoo if both import and export work fine
* It wouldn't be correct to categorize this model as `detection`, we can instead create a category `misc` and place it under that
* All add this to `https://github.com/Cloud-CV/Fabrik/blob/master/tutorials/tested_models.md`

gautamjajoo:
@utsavgarg So, what should I do so that it exports successfully?

utsavgarg:
You will have to identify what is causing the problem and if it's a minor one just make the fix or let us know, we will fix the error.

gautamjajoo:
@utsavgarg I could not find the error. Also, we can add the DEX-Age Model but there is same error of exporting.

utsavgarg:
Closing as export for this doesn't work

PR #220, opened 2017-12-28
Fix: numeric values should be interpreted as numbers

NSDCars5:
This was referenced in the discussion of #210. A numeric field in a layer's properties, when changed, would be interpreted as a string, which would cause export of a modified layer to fail. Now numeric fields are converted to a number before being added into the net.


PalashTanejaPro:
I have fixed this issue on the backend side in my PR.

utsavgarg:
LGTM :)

PR #221, opened 2017-12-28
Added DeepYeast and SpeechNet to model zoo

adamstafa:
Caffe export for both works fine, Keras and Tensorflow fail.

### DIFF
--- a/example/caffe/DeepYeast.prototxt
+++ b/example/caffe/DeepYeast.prototxt
@@ -1,4 +1,4 @@
-name: "HOwt_png_quick"
+name: "DeepYeast"
 layer {
   name: "control"
   type: "Data"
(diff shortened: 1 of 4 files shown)


example/caffe/DeepYeast.prototxt
```diff
@@ -0,0 +1,412 @@
+name: "HOwt_png_quick"
```
utsavgarg: Change this to the name of the model

adamstafa:
I converted old prototxt of [SpeechNet](https://github.com/pannous/caffe-speech-recognition) to new one.   Caffe export and import works great but the `param { lr_mult: 1 decay_mult: 1 }` is converted to only `{ lr_mult:1 }` - thus `decay_mult` is lost.

utsavgarg:
Closing is as this is similar to other samples present in the model zoo.

PR #225, opened 2017-12-29
Pass errors to UI for tensorflow and Keras import error.

trexdinosaur:


trexdinosaur:
@utsavgarg is the coverage decrease fine?

I can submit a pr for more tests in a separate PR if needed.

utsavgarg:
@trexdinosaur Yes, a slight coverage decrease should be alright. Also, are there any specific json/pbtxt where this behaviour was observed?

PR #226, opened 2017-12-30
More detailed error messages

adamstafa:
Now it show why is prototxt invalid and unknown layers.


utsavgarg:
This PR adds more informative error messages for Caffe import and Keras export.

utsavgarg:
LGTM :+1:

Issue #223, opened 2017-12-29
Press button display with and without adding layer.


Obtin:
- the expected behaviour
When no layer is added the Press button should indicate its that it is non-functional.
- the observed behaviour
The button appears the same when being functional (layers added) and even when it is not.(no layers added).

Thanks!
when no layer added.
![fabrik1](https://user-images.githubusercontent.com/25207205/34439742-18abd4b0-ecd6-11e7-867b-959556423a5e.png)
after adding layers
![fabrik2](https://user-images.githubusercontent.com/25207205/34439747-20ab28dc-ecd6-11e7-8bec-8228fe25b3f2.png)

nu11us:
I'm not sure what the problem is. When clicked on a button, it simply adds itself to the bottom of the network. If there is no layer bottom to be connected, it will not work. Are you suggesting that if it is not working it should be grayed out? If there is desire for that, I can easily edit the script for that. 

**Image Data**, the first layer that you tested, cannot be added to itself, as it has only output. Since there is an output, but no input, the button will not work. If you were to dress **Image Data** then something like **Convolution**, which has both input and output, it will connect.

Obtin:
@nu11us, when no layer is added the Press button should act as it is deactivated. 
![2](https://user-images.githubusercontent.com/25207205/34450899-47ffcc46-ed3d-11e7-9361-67a9879f409a.png)

And after adding layers when it is enabled, it may appear like this.
![1](https://user-images.githubusercontent.com/25207205/34450900-4b1733ec-ed3d-11e7-8b48-0eef9c8f843f.png)

nu11us:
But what if someone wants to use the layer through dragging? Each button has two main abilities: it attaches to the bottom of the graph when it is clicked, and it is placed when it is dragged and dropped. What you are suggesting would deactivate the button that could still be dragged.

Obtin:
I am very sorry @nu11us I did not get you. Please could you explain to me what does the button do when no layers are added and the nothing to work on.
![image](https://user-images.githubusercontent.com/25207205/34456019-f3fe5ea8-edb1-11e7-830c-6ae050c0ba94.png)
Thank you. :-)

nu11us:
Sorry about that. I thought you were referring to one of the buttons on the left. The press buttons should be disabled when then are no layers. I thought you meant once a button (i.e. Input Data) was pressed to create a layer. Yeah, *Press ]* and *Press [* should be disabled. Will get to that shorty.

[...]

PR #228, opened 2017-12-30
Disable buttons when unusable

nu11us:
Refer to issue #223

### DIFF
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -20,6 +20,7 @@ class Canvas extends React.Component {
     this.hover = 0;
     this.mouseState = null;
     this.placeholder = true;
+    this.disableZoom = true;
   }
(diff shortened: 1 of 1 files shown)


utsavgarg:
@nu11us A few problems I saw on testing:
* Even though the button looks disabled, clicking on it is changing the zoom levels, you can notice this after you click on the disabled buttons and then load a net.
* The button remains disabled when you add a single layer by clicking it, it only became enabled when you drag and drop a layer or import a net.

utsavgarg:
@nu11us are you still working on this?

nu11us:
@utsavgarg Yeah, I am, but the tasks layer is taking up my time. [Planned for this task](https://codein.withgoogle.com/tasks/6548969301737472/?sp-organization=4854733644759040&sp-is_beginner=False)

PR #232, opened 2018-01-01
Added a New Model LeNet

gautamjajoo:
Added a new "LeNet" Model and removed the Dex-Gender and Dex-Age Model and export is also working fine.
Thanks! @utsavgarg

PR #233, opened 2018-01-01
Update tested_models.md

gautamjajoo:
With reference to #232 
Thanks! @utsavgarg


utsavgarg:
Please push this commit to the same PR.

PR #234, opened 2018-01-02
Added a New Model - Cifar10

gautamjajoo:
Added a New Model - Cifar10. 
CIFAR-10 is an established computer-vision dataset used for object recognition.
Thanks! @utsavgarg


utsavgarg:
Closing as this is very similar to other models present in the model zoo.

PR #237, opened 2018-01-03
added cifar10_cnn model from caffe to zoo

nchhaj189:


utsavgarg:
Hi, the export to Keras is not working, in the future check that export to at least the framework from which you are importing works. Also, add the model to `tested_models` list in the tutorials folder and add a link to the source from where you got this network.

virajprabhu:
Thanks @nchhaj189. Can you modify the PR name to the right model?

virajprabhu:
The correct title should also be displayed in the model zoo dashboard, and in the title pane on the canvas.

virajprabhu:
image

utsavgarg:
LGTM :)

PR #239, opened 2018-01-03
Fix: Update Total Parameter for layer add and delete(#126)

### LINKED ISSUE #126: Display parameters of the model being visualized

Ram81:
@utsavgarg Fixes #126

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -323,10 +323,13 @@ class Content extends React.Component {
   }
   calculateParameters(net) {
     // Iterate over model's each layer & separately add the contribution of each layer
+    var totalParameters = 0;
     Object.keys(net).sort().forEach(layerId => {
       const layer = net[layerId];
-      this.updateParameters(layer, net);
+      net[layerId]['shape']['parameters'] = this.getLayerParameters(layer, net);
+      totalParameters += net[layerId]['shape']['parameters'];
     });
+    this.setState({ net: net, totalParameters: totalParameters});
   }
   loadLayerShapes() {
     this.dismissAllErrors();
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -77,7 +77,8 @@ class Content extends React.Component {
     this.setModelFramework = this.setModelFramework.bind(this);
     this.loadLayerShapes = this.loadLayerShapes.bind(this);
     this.calculateParameters = this.calculateParameters.bind(this);
-    this.updateParameters = this.updateParameters.bind(this);
+    this.getLayerParameters = this.getLayerParameters.bind(this);
+    this.updateLayerShape = this.updateLayerShape.bind(this);
     this.modalContent = null;
     this.modalHeader = null;
     // Might need to improve the logic of clickEvent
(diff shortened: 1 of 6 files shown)


utsavgarg:
@Ram81 this looks good but the calculation doesn't seem to be working when adding a new layer, this might be because it's not connected when added and the input and output shapes are not defined. We might need to trigger the calculation when the parameter values are changed rather than when a layer is added.

Ram81:
@utsavgarg Yeah I also noticed that, do you want me to add the logic for modification of parameters and connection establishment updates in same PR ?
Because I thought you would want that handled separately for a new connection addition & modification of parameters, as new connection could be added in between too i.e. non-leaf nodes of graph.

utsavgarg:
@Ram81 Nice effort :) It's definitely an improvement from before, but I still see some issues with the calculation
ezgif com-video-to-gif

ide/static/js/content.js
```diff
@@ -92,8 +93,28 @@ class Content extends React.Component {
...
+    net[layerId] = layer;
+    // Parsing for integer parameters when new layers are added as by default all params are string
+    var intParams = ["crop_size", "num_output", "new_height", "new_width", "height", "width", "kernel_h", "kernel_w",
+                      "kernel_d", "stride_h", "stride_w", "stride_d", "pad_h", "pad_w", "pad_d", "size_h", "size_w",
+                      "size_d", "n"];
+    Object.keys(net[layerId].params).forEach(param => {
+      if (intParams.includes(param)){
+        net[layerId].params[param][0] = parseInt(net[layerId].params[param][0]);
+        if (isNaN(net[layerId].params[param][0]))
+          net[layerId].params[param][0] = parseInt(0);
```
utsavgarg: is a `parseInt` necessary here?

ide/static/js/content.js
```diff
@@ -92,8 +93,28 @@ class Content extends React.Component {
   }
   addNewLayer(layer) {
     const net = this.state.net;
-    net[`l${this.state.nextLayerId}`] = layer;
-    this.setState({ net, nextLayerId: this.state.nextLayerId + 1 });
+    const layerId = `l${this.state.nextLayerId}`;
+    var totalParameters = this.state.totalParameters;
+    net[layerId] = layer;
+    // Parsing for integer parameters when new layers are added as by default all params are string
```
utsavgarg: Also add a comment that some parameters might have been missed, or are you sure you've covered all?

ide/static/js/content.js
```diff
@@ -92,8 +93,28 @@ class Content extends React.Component {
...
+                      "kernel_d", "stride_h", "stride_w", "stride_d", "pad_h", "pad_w", "pad_d", "size_h", "size_w",
+                      "size_d", "n"];
+    Object.keys(net[layerId].params).forEach(param => {
+      if (intParams.includes(param)){
+        net[layerId].params[param][0] = parseInt(net[layerId].params[param][0]);
+        if (isNaN(net[layerId].params[param][0]))
+          net[layerId].params[param][0] = parseInt(0);
+      }
+    });
+    //console.log(layer);
```
utsavgarg: Remove

ide/static/js/content.js
```diff
@@ -92,8 +93,28 @@ class Content extends React.Component {
...
+    this.updateLayerShape(net, layerId);
+    // Check for only layers with valid shape
+    if (net[layerId]['shape']['input'] != null && net[layerId]['shape']['output'] != null) {
+      net[layerId]['shape']['parameters'] = this.getLayerParameters(net[layerId], net);
```
utsavgarg: It's probably better to store this in `info` rather than `shape`

ide/static/js/content.js
```diff
@@ -122,8 +143,16 @@ class Content extends React.Component {
 
   modifyLayer(layer, layerId = this.state.selectedLayer) {
     const net = this.state.net;
+    var oldLayerParams = this.state.totalParameters;
```
utsavgarg: Instead of repeating this portion of the code from `modifyLayerParams` can we have a separate function which has this in one place and both these functions can call it.

ide/static/js/content.js
```diff
@@ -202,11 +233,43 @@ class Content extends React.Component {
...
+      url: 'layer_parameter/',
+      dataType: 'json',
+      type: 'POST',
+      async: false,
+      data: {
+        net: JSON.stringify(netData),
+        layerId: layerId
+      },
+      success : function (response) {
+        //this.setState({ net: response.net, nextLayerId: this.state.nextLayerId + 1 });
```
utsavgarg: Remove

ide/static/js/setParams.js
```diff
@@ -22,10 +22,19 @@ class SetParams extends React.Component {
     const net = this.props.net;
     let layer = net[this.props.selectedLayer];
     layer = JSON.parse(JSON.stringify(layer));
+    var intParams = ["crop_size", "num_output", "new_height", "new_width", "height", "width", "kernel_h", "kernel_w",
```
utsavgarg: Do we need to parse these as `int` at two places?

Ram81: Yeah because when layer is added it doesn't use setparams

ide/static/js/content.js
Ram81: I modified old logic where we have to update layer after param pane is closed which does exactly the same thing you are asking for, now we don't call modifyLayer from modifyLayerParams

PR #244, opened 2018-01-04
Zoo data change

trexdinosaur:
Zoo Modal now pulls models from a JSON dataset, makes it easier to adjust in the future rather than changing all the ModelElements!


trexdinosaur:
@utsavgarg once you give me the OK that this change is fine, I'll resolve conflicts!

trexdinosaur:
Also, I think we should implement a similar change for ```pane.js``` as the code is very repetitive right now, and it should be Reactified :)

utsavgarg:
@trexdinosaur The idea is nice but there seems to be some issue with the implementation, `webpack` throws some errors while building this.

PR #264, opened 2018-01-10
Import model via text input

NSDCars5:
For Code-In task: https://codein.withgoogle.com/dashboard/task-instances/5870321804509184/

- [x] Create a text editor which pops up as a modal with tabs at the top to choose between Caffe/Keras/Tensorflow and let the user directly paste the model config into this editor.

- [x] There should be a submit button at the bottom upon clicking which the model should be loaded. 

- [x] If the user launches the editor again after submitting it, the old config should still be in there.

### DIFF
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -544,6 +544,7 @@ layer_dict = {'Accuracy': Accuracy,
 
 @csrf_exempt
 def import_prototxt(request):
+    prototxtIsText = False
     if request.method == 'POST':
         if ('file' in request.FILES) and \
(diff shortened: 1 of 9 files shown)


utsavgarg:
A few UI changes:
* The buttons in the top bar need to be spaced out evenly, the current version doesn't look good
image
* The heading of the modal should be styled similar to the headings in the model zoo
* The tabs at the top should be styled similar to the train/test phase tabs and make Caffe selected by default (except maybe the colour scheme)
* The modal itself should not be scrollable, 
image
Only the text input area should be scrollable, also style this scrollbar similar to the other ones in the application.

NSDCars5:
@utsavgarg Made the requested changes.

PR #267, opened 2018-01-10
added flickr caption model

nchhaj189:
added flickr image caption model


utsavgarg:
Closing as this is not an image captioning model.

PR #266, opened 2018-01-10
Login

trexdinosaur:
Adds authentication and ability to store models
Also - allows Caffe_App to get migrations... Looks like these got removed accidentally by somebody. 

Longer timeframe (3-4 days):
- Versioning for the models
- User based authentication for a model VS just a URL/ID (think: shareable link in Google Drive VS adding a person...) (@utsavgarg is this something you are looking for? should be very easy to implement.)
- tests for all of this (these might take me awhile... :) )

PR #269, opened 2018-01-11
Fix and commonize export prep

NSDCars5:
This is parallel to #190, for the share link button instead of the export button.

For now I have made the same change here, but maybe we could look into separating this code into its own function and call saveDb() and exportNet() as a callback, so we don't have to maintain the code in two places?

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -693,14 +693,14 @@ class Content extends React.Component {
   saveDb(){
     this.dismissAllErrors();
     const error = [];
-    const net = this.state.net;
+    const net = JSON.parse(JSON.stringify(this.state.net));
 
     Object.keys(net).forEach(layerId => {
       const layer = net[layerId];
       Object.keys(layer.params).forEach(param => {
         layer.params[param] = layer.params[param][0];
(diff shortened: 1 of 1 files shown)


trexdinosaur:
@NSDCars5 a lot of this is being refactored with #266, I can work on getting saveDB and exportNet working better together

NSDCars5:
@trexdinosaur If that PR is ready to merge, I could just PR the required commits into your branch? I have the changes ready.

trexdinosaur:
@NSDCars5 you have the changes for saveDB and exportNet? 
I'm not done with the changes with the login, but I don't think there will be many changes to ```content.js```, so you can pr to my ```login2``` branch...

utsavgarg:
@NSDCars5 Thanks, yes we could do this at one place instead of maintaining the code in two places. Also, it would be better to do it here instead of merging it with @trexdinosaur's PR as it's easier to keep track.

NSDCars5:
Sorry about the close/reopen. I'll put the second commit here when I get home. @utsavgarg

NSDCars5:
@utsavgarg Made the new changes.

PR #276, opened 2018-01-13
Load model from url

adamstafa:
Works similarly as #264.

### DIFF
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -6,6 +6,7 @@ from caffe.proto import caffe_pb2
 from google.protobuf import text_format
 import tempfile
 import subprocess
+import urllib2
 
 
 # ******Data Layers******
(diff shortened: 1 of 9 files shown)


utsavgarg:
Nice work :) The functionality seems to be working well, a few changes:
* Instead of putting a separate button for this in the top bar, put it as an element in the import dropdown
* To import anything from GitHub we need to get the link for the `raw` file and then paste it, if we paste the link for the page instead it downloads the HTML and the error message doesn't help, two options:
        - Find a way to directly download file from their page e.g. `https://github.com/Cloud-CV/Fabrik/blob/master/example/caffe/GoogleNet.prototxt`
        - Have a better error message which explains the issue and gives the solution

adamstafa:
@utsavgarg Made the requested changes. Shall I write missing unit tests to increase coverage?

utsavgarg:
@adamstafa Yes, that would be great

PR #283, opened 2018-01-16
Add UNET to The Model Zoo

galiaxy:
I added [UNET](https://github.com/ZFTurbo/ZF_UNET_224_Pretrained_Model) to the Model Zoo, this is a screenshot:
unet model
I've also made sure that I can export it successfully.


gautamjajoo:
@galiaxy I think so Utsav sir told that they are only adding popular models. WHile you can add to tested models.md .

galiaxy:
@gautamjajoo Right, although I  think UNet is a pretty popular model, except this UNet I added here, is a modification - so yes, I've been thinking about that too. Well, should I try another model then?

gautamjajoo:
@galiaxy So you can try Neuraltalk2

galiaxy:
@gautamjajoo True - but it uses Torch so I should convert it first

virajprabhu:
LGTM, thanks @galiaxy!

PR #287, opened 2018-01-18
LRN support for Tensorflow

adamstafa:
`Tensorflow export` has been using `Keras export` to make a JSON and then it converts this JSON to pbtxt. Problem is that Keras doesn't support `LRN` layer (TF does) so I added a custom one. The `LRN` layer is added to the `layer_map` in [export_json.py](https://github.com/Cloud-CV/Fabrik/blob/master/keras_app/views/export_json.py) only when exporting to TF thus it cannot be used in Keras export.

I successfully tested it on `AlexNet` that was failing on the `LRN` layer. There weren't any conflicts with other existing nets.


utsavgarg:
@adamstafa This is really nice :) If the custom layer is written in Keras it should be possible to use it for Keras export as well. Put the custom layer in a separate folder (maybe outside views), if any exported file uses the custom layer download it as well so that the user can use it and also test this for GoogLeNet currently the export is failing.

adamstafa:
@utsavgarg I moved the layer to `keras_app/custom_layers` and also added it for Keras export.

But I don't think that the layer should be downloaded automatically. It would be kinda annoying when you export a net multiple times. I would add a modal with a link to download the layer instead.

PalashTanejaPro:
@adamstafa @utsavgarg 
How about a popup on detecting a custom layer?  
That'd be more user friendly.

adamstafa:
@PalashTanejaPro Do you mean something like the error popup?

PalashTanejaPro:
@adamstafa Yeah something like that but  green color and which allows the user to click on it to download.

adamstafa:
@PalashTanejaPro Sure, that would be great! But the layers should be IMO detected in backend and sent to the frontend. Frontend should only display these layers.

PalashTanejaPro:
@adamstafa Aren't error also sent from the backend though. Also let's move this chat to gitter

utsavgarg:
@adamstafa I am getting an error when trying to export GoogLeNet to both Keras and TF, can you look into it?

adamstafa:
@utsavgarg I'm getting the error too, but it isn't caused by the LRN layers.

ide/static/js/content.js
```diff
@@ -411,6 +414,27 @@ class Content extends React.Component {
             downloadAnchor.download = response.name;
             downloadAnchor.href = response.url;
             downloadAnchor.click();
+            if ('customLayers' in response && response.customLayers.length !== 0) {
+              this.addInfo(
+                <span>
```
utsavgarg: Can we handle this formatting in the `info` component instead of putting it in `content.js`.

keras_app/custom_layers/lrn.py
```diff
@@ -0,0 +1,40 @@
+from keras.layers.core import Layer
```
utsavgarg: Add from where this implementation was taken.

keras_app/startup.py
```diff
@@ -0,0 +1,11 @@
+import os
+from custom_layers import config
+
+
+def run():
+    BASE_DIR = os.path.dirname(
```
utsavgarg: Could move this outside the function

tensorflow_app/views/json2pbtxt.py
```diff
@@ -26,7 +27,8 @@
     json_str = f.read()
 
 json_str = json_str.strip("'<>() ").replace('\'', '\"')
-model = model_from_json(json_str)
+lrn = imp.load_source('LRN', BASE_DIR + '/keras_app/custom_layers/lrn.py')
+model = model_from_json(json_str, {'LRN': lrn.LRN})
```
utsavgarg: Why do we always pass LRN? Might be better to do so if the layer exists.

keras_app/views/layers_export.py
```diff
@@ -642,6 +643,16 @@ def time_distributed(layerId, idNext, net, layer_in, layer_map):
     out = {}
     out[layerId] = TimeDistributed(
         layer_map[net[idNext]['info']['type']](net[idNext], layer_in, idNext, False)[idNext])(*layer_in)
+
+
+# LRN for Tensorflow export only
```
utsavgarg: Seeing that we use the Keras backend to export TF, and the custom layer is written in Keras, I feel there should be a way to use it for Keras export as well. Can you check this once?

ide/static/css/style.css
```diff
@@ -694,6 +694,30 @@ button.url-import-modal-button {
     white-space: pre-line;
 }
 
+.info{
+    padding: 10px;
+    width:40%;
+    padding-left: 10px;
+    color: white;
+    background-color: #2472c1;
```
utsavgarg: Change this to a shade of green.

utsavgarg:
@adamstafa Have you been able to look the changes?

keras_app/views/layers_export.py
adamstafa: In the latest commit it's possible to use it with Keras too. I just forgot to delete that comment.

adamstafa:
@utsavgarg I was kinda busy lately (and probably will be) so I don't have time to do that. Could you assign it to someone else please?

utsavgarg:
Sure

adamstafa:
Guess we're done here, ni🅱️🅱️as. Thanks @Ram81 for finishing my work 👍

Issue #293, opened 2018-01-28
Extend Tensorflow support


utsavgarg:
For Tensorflow, the import feature currently only supports a few layers (~10), we want to extend this to support as many layers as are available in our Keras backend (~60). Also, the current export for Tensorflow uses the Keras backend, we want to write a separate backend for Tensorflow export so that backends are not as highly coupled.

Ram81:
@utsavgarg I would like to work on this one

Shivanshmundra:
@Ram81 What is the progress on this issue, if not much I would like to take this issue.

Ram81:
@Shivanshmundra sorry but I am working on this, please take up some other issue.

Shivanshmundra:
Ok, I will look up for some another issue.

PR #296, opened 2018-01-29
Fix: Extended Support for Layer Import for  TF and keras (#293)

Ram81:
@utsavgarg please review.

### DIFF
--- a/caffe_app/views/import_prototxt.py
+++ b/caffe_app/views/import_prototxt.py
@@ -469,6 +469,16 @@ def ContrastiveLoss(layer):
     return params
 
 
+def Concat(layer):
+    params = {}
+    if (layer.concat_param.axis is not None):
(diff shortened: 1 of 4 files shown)


caffe_app/views/import_prototxt.py
```diff
@@ -469,6 +469,16 @@ def ContrastiveLoss(layer):
     return params
 
 
+def Concat(layer):
+    params = {}
+    if (layer.concat_param.axis is not None):
+        params['axis'] = layer.concat_param.axis
+    else:
+        # default value for axis of dropout in caffe
```
utsavgarg: How is this related to dropout?

Ram81: Sorry I accidentally wrote dropout there it should be concat

PR #297, opened 2018-02-01
Fix: Corrected cmake option for caffe installation and a broken link

thatbrguy:
In caffe_tensorflow_keras_install.sh, the correct cmake option to build the python layer is:
`-DBUILD_python_layer=1`
Whereas, the existing option was:
`-DWITH_PYTHON_LAYER=1`

Which is a make option, and throws the following warning with cmake:
`CMake Warning:
Manually-specified variables were not used by the project:
WITH_PYTHON_LAYER`

By default, python layers is enabled in CMakeLists.txt, and hence it gets ignored. But for sanity, I've updated the change in the script.


thatbrguy:
Fixed a broken link in the Readme file.

PR #299, opened 2018-02-04
Fix: Add check for layer shape in parameter calculation

Ram81:
@utsavgarg minor fixes to avoid unnecessary error while fetching total number of parameters of a layer.

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -342,7 +342,7 @@ class Content extends React.Component {
   }
   loadLayerShapes() {
     this.dismissAllErrors();
-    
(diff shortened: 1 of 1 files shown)


thatbrguy:
@Ram81 I think it's better to throw an Exception instead, due to issues like #300.

Ram81:
@thatbrguy It's not an issue with the model user is creating its a measure related to shape calculation, whenever user updates parameters of a layer shape recalculation logic is triggered which is why if there is no dimensions available for input layer the layers following it will not have a shape, the shape calculation is under development and we are still working on ways to achieve the changes in shape for model being built.

PR #304, opened 2018-02-13
Fix: 1D padding logic in layers_import.py

thatbrguy:
In keras_app/views/**layers_import.py**, function **get_padding()** originally had:

```
def get_padding(params, input_shape, output_shape, pad_type, type):
    k_w, k_h, k_d, s_w, s_h, s_d = params
    if (type == '1D'):
        if (pad_type == 'valid'):
            return 0
        else:
            pad_h = ((output_shape[2] - 1) * s_h + k_h - input_shape[2]) / 2
            return pad_h
```

The calls for get_padding() in 1D Conv and 1D Pool provided the **params** argument as:
```
get_padding([params['kernel_w'], -1, -1, params['stride_w'], -1, -1],
                                          layer.input_shape, layer.output_shape,
                                          padding, '1D')
``` 
Where only **k_w** and **s_w** are defined. Hence, we need to calculate **pad_w** and not pad_h. Modified code to calculate pad_w.


utsavgarg:
LGTM :) Thanks

PR #305, opened 2018-02-13
TF import support

Ram81:
@utsavgarg minor correction in batchnorm layer now resnet shape calculation works fine and support leakyrelu as we are using same frontend for leakyrelu and relu.


utsavgarg:
LGTM :) Thanks

PR #307, opened 2018-02-14
Fix #306:changed shellscript to append new entries into `bash_profile` instead of wiping and rewriting the file

### LINKED ISSUE #306: installation shellscript wipes present entries from .bash_profile
Present shellscript is pretty good and makes source installation pretty smooth. The only problem I noticed is that it wiped my present PATH entries from `.bash_profile` file while installing caffe py-package in this line https://github.com/Cloud-CV/Fabrik/blob/master/requirements/caffe_tensorflow_keras_install.sh#L30 . I have made a fix, would push it soon.

kapilkd13:
fixes #306 

shellscript wipe present entries in bash_profile while setting PYTHONPATH, modifying it to only append at the end of bash_profile instead of deleting present entries.

### DIFF
--- a/requirements/caffe_tensorflow_keras_install.sh
+++ b/requirements/caffe_tensorflow_keras_install.sh
@@ -27,7 +27,7 @@ if [ ! -d $HOME/caffe/caffe ]; then
 		cmake -DCPU_ONLY=1 -DBUILD_python_layer=1 ..
 		make -j"$(nproc)"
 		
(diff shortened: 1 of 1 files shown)


deshraj:
Sorry but I dont think we can accept this change since if you only use a single `>` you will completely overwrite the contents of the file.

kapilkd13:
@deshraj Sorry I didn't got your point. Are you suggesting the that the present use of single `>` is intended.
Just to clarify, presently we are using `>` this PR changes it to `>>`.

deshraj:
Oops my bad. I somehow saw it the reverse way. Reopening the pull request.

PR #309, opened 2018-02-15
Support and unittests for Keras using TF backend 

thatbrguy:
**Change Log:**
- Fixed a failing unittest while using TensorFlow backend for Keras.
- Added support for SeparableConv2D, which was not possible with Theano backend.
- Added unittest for Import and Export of SeparableConv2D. These unittests are **skipped** if backend used is Theano.
- Added unittest for testing dimension ordering (channels last/first) . By default, in our case, the channels last ordering is used for both Theano and TensorFlow. The unittest is used to verify the same.
- Modified all import unittests to use channels last ordering, as it should be. The export tests only tests if the layer's name is correct, so it was left intact.
- Added a new test entry for SeparableConv2D (DepthwiseConv) in keras_export_test.json
- Modified README to change server deploying instruction to use the default backend (tensorflow). 

@utsavgarg please review.


utsavgarg:
@thatbrguy Can you move the addition of `SeparableConv2D` to a separate PR and let this be just for shifting the Keras backend from Theano to TF. Also, the build is failing, can you fix that?

thatbrguy:
@utsavgarg I'll close this PR and open new ones addressing both separately.

PR #310, opened 2018-02-18
Keras TF Backend Support

thatbrguy:
**Update:**

- Added an extra dependency for TensorFlow installation, and exported a variable to fix the build error.

**Change Log:** (Only the TF Backend part of #309)

- Fixed a failing unittest while using TensorFlow backend for Keras.
- Added code to reset graph before importing, in tensorflow_app/views/**import_graphdef.py**
- Added unittest for testing dimension ordering (channels last/first) . By default, in our case, the channels last ordering is used for both Theano and TensorFlow. The unittest is used to verify the same.
- Modified all import unittests, and a line in import_graphdef.py, to use channels last ordering, as it should be. The export tests only tests if the layer's name is correct, so it was left intact.
- Modified README to change server deploying instruction to use the default backend (TensorFlow). 
- Modified travis.yml to run tests with keras backend as TensorFlow

@utsavgarg please review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ install:
 
 before_script:
   - export PYTHONPATH=$PYTHONPATH:$HOME/caffe/caffe/python
+  - export LD_PRELOAD="/usr/lib/libtcmalloc.so.4"
   - psql -c "CREATE DATABASE fabrik" -U postgres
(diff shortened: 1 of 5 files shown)


jahsrasta:
Very good, great support, thank you, God bless.

PR #313, opened 2018-02-24
Keras Separable Conv2D with TF Backend

thatbrguy:
**Change Log:** (The 2nd part of #309)

- Separable Conv2D was only compatible with TF backend and not Theano. Since master is in sync with TF backend, I'm introducing functionality for SeparableConv2D.
- Wrote the missing export unittest for the same. The import and export unittests are skipped if backend used is theano.

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -282,7 +282,7 @@ class Content extends React.Component {
     var weight_params = 0;
     var bias_params = 0;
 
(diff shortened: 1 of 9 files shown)


thatbrguy:
@utsavgarg I'll account for these mistakes soon

Ram81:
@thatbrguy support for depthwise conv was added in #354. Closing this.

PR #319, opened 2018-03-01
Fix #318: fixed datatype error

### LINKED ISSUE #318: "str" error while editing some params value in layers
While editing some parameters value in layer it throws an error when exporting the model:
` "(edited float value)" has type str, but expected one of: int, long, float`

error_float

Shivanshmundra:
Fix #318: Fixed datatype error 
This fixes the issue #318 Don't throw error when some specific parameters value is edited while model is exported. Checked by again importing the same model showing same values of edited parameters.

### DIFF
--- a/ide/static/js/data.js
+++ b/ide/static/js/data.js
@@ -59,7 +59,7 @@ export default {
       scale: {
         name: 'Scale',
         value: 1.0,
-        type: 'float',
+        type: 'number',
         required: false
       },
       mirror: {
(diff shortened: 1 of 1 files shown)


utsavgarg:
@Shivanshmundra Thanks for reporting the bug and the fix. But the error seems to persist even after this. Steps to recreate:
* Load ResNet-101 model from the model zoo
* Change value of `moving average fraction` in the batch norm layer
* Try exporting

Ram81:
@Shivanshmundra it would be better to parse values at the time of layer update. If you need help with update part we can discuss.

Shivanshmundra:
@utsavgarg @Ram81 I figured out the error, Actually by default each value is parsed as `string`.
Now there is exception to parse value as `integer` but there was no exception for `float` so I added it and now its working fine.
Have a review.

Issue #311, opened 2018-02-23
Modify model zoo UI


utsavgarg:
The model zoo currently opens up as a modal, and therefore cannot hold all models that have been [tested with Fabrik](https://github.com/Cloud-CV/Fabrik/blob/master/tutorials/tested_models.md) as it will get too cluttered. Even without all the models present, there are many models which aren't as popular and it might be confusing for new users who might not be familiar with them.

Therefore, it might be better to initially show only popular models(which have well-cited research papers) in the zoo and have a *load more* button at the bottom which reveals all the remaining models.

The task has the following deliverables:
- [ ] Add all models from [tested models](https://github.com/Cloud-CV/Fabrik/blob/master/tutorials/tested_models.md) to the model zoo
- [ ] Only show popular models initially
- [ ] Reveal all remaining models on clicking the *load more* button

Extended functionality:
- [ ] Redirect to a new page when the user clicks *load more*, the new page has a search bar at the top
- [ ] The new page should have models displayed in card view each with a preview of the architecture

thatbrguy:
@utsavgarg With reference to an old issue #300, the links at tested_models.md needs to be replaced with newer share links, to fix a bug related to parameter count and keras export. I would like to suggest that this be added as a sub-issue for anyone who wants to take it.

MuLx10:
I want to take up this issue.I am currently working on it

laylalaisy:
Do you think it's proper and better if we add an search bar at the top of the model ZOO UI,  so the user can search the wanted model quicker? @utsavgarg

utsavgarg:
@laylalaisy Yes, that is a good idea. We can include a search bar which dynamically filters the list of models.

laylalaisy:
@utsavgarg Thx and I would like to implement it! I think search bar can be more convenient for some users.

utsavgarg:
@laylalaisy it's a nice effort, but the UI doesn't seem to scale well with different screen sizes. Also, I have updated the feature description, see if you can incorporate those changes in this PR.

![image](https://user-images.githubusercontent.com/11680179/37117377-03e2b422-228c-11e8-9992-77665e4271d3.png)

utsavgarg:
@MuLx10 @laylalaisy Instead of both of you working on this, it would be better for either one of you to take the issue or split the features so that no effort is wasted.

MuLx10:
@utsavgarg I'll incorporate the extensions and discuss the split up with @laylalaisy

PR #321, opened 2018-03-02
Fix #311 : Modifies model zoo UI 

laylalaisy:
Fix #311: added a search bar which dynamically filters the whole list of models and directly load them by clicking.
This adds the issue #311. In the model zoo, users can search models and related options will be listed below. Then, users can directly click the option to load the model. Besides, uppercase/lowercase/fixed and any part of models' name are available to be searched. If users do not search in the search bar, frequently-used models will be listed below as recommendation. (PS: this fix is based on the work from MuLx10 #315 )

PR #315, opened 2018-02-25
Fix #311 : Modifies model zoo UI

MuLx10:
Fix #311  : Added all the tested models. Added load more button.
This fixes the issue #311 show popular models initially in model zoo ,reveals all remaining models on clicking the load more button , also added all models from tested models to the model zoo.

PR #325, opened 2018-03-10
Modify model zoo UI #311

MuLx10:
Fixes issue #311 
This PR incorporates a load more button which redirects to the zooSearch page. The zooSearch page displays the models in the card view with a preview of the architecture.


Ram81:
@MuLx10 I think you can remove package.json from the PR.

utsavgarg:
@MuLx10 If this covers changes done in https://github.com/Cloud-CV/Fabrik/pull/315 can you close that?

utsavgarg:
@MuLx10 This is really nice :) A few suggestions:
image
The text in the zoo isn't centred, there seems to be more whitespace on the right.

Also, by redirecting to a new page, I meant a fully different view, not in a modal, as this seems cluttered. So if we can have a new page, which shows all models arranged in card view by category and have a search bar at the top which filters the models dynamically, that would be great.

utsavgarg:
Also, if this covers the changes made in #321, please close that too.

laylalaisy:
Ok，I will check that tomorrow. Since it’s midnight in China and I am in gym now. Thank you and I’m looking forward to making more contributions to Fabrik. I definitely like this project!

PR #329, opened 2018-03-13
Fix #317:  Added feature for multiple deletion of layers

### LINKED ISSUE #317: Feature Request: Delete multiple layers in Fabrik
As of now, we can delete one layer at a time. We have to select the layer we want to delete and scroll all the way to the bottom to delete the layer. 

It's usually a hindrance if you want to make lots of changes to your network, but also do not want to clear everything.

A good solution would be to introduce an option to select and delete multiple layers. Please be sure to update parameter count and other relevant attributes accordingly.

Shivanshmundra:
Fix #317: Added feature for multiple deletion of layers
I added two buttons to include this feature:
* **Select Multiple Layer** for choosing layers to delete.
* **Delete Selected Layers** for deleting the selected layers.

I am working on the issue of stopping navigation drawer while layers are selected.

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -41,11 +41,13 @@ class Content extends React.Component {
       rebuildNet: false,
       selectedPhase: 0,
       error: [],
+      multiple_layerId: [],
       load: false,
       modalIsOpen: false,
       totalParameters: 0,
       modelConfig: null,
       modelFramework: 'caffe'
+      
     };
     this.addNewLayer = this.addNewLayer.bind(this);
     this.changeSelectedLayer = this.changeSelectedLayer.bind(this);
@@ -89,6 +91,8 @@ class Content extends React.Component {
     // Might need to improve the logic of clickEvent
     this.clickEvent = false;
     this.handleClick = this.handleClick.bind(this);
+    this.store_list_layers = this.store_list_layers.bind(this);
+    this.delete_selected_layers = this.delete_selected_layers.bind(this);
   }
   openModal() {
     this.setState({modalIsOpen: true});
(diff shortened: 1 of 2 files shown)


Ram81:
@Shivanshmundra can you share a screenshot?

Shivanshmundra:
Yeah Sure:
multiple_layer_deletion

@Ram81  Is this what you wanted?

Ram81:
@utsavgarg does this design for deleting multiple layers looks good to you?
I have some alternatives in mind if needed

thatbrguy:
My two cents would be that, as mentioned in #317, I think a "delete mode" button in the actions tab would be a bit neater.

Shivanshmundra:
@thatbrguy I didn't understand clearly what you said #317, like adding a delete mode and then selecting the layers to delete.I think above solution does that just in another way if I am getting you right.
Can you please elaborate what you want?Also if @Ram81 have any good suggestions.

Ram81:
@Shivanshmundra Delete multiple layers is not a functionality of layer, you can provide a delete multiple layer button in zoo tab under some tab say 'Action'  which will allow user to choose multiple layers and then use a delete action to delete all of them at once. 
It is not a feature that should be available in the layer parameters pane as per my views. Try to think where is the most suitable place to put the 'Delete multiple layer' action.

utsavgarg:
@Shivanshmundra Thanks for the effort. The choice of placing the delete multiple layer button in each layer pane is a bit odd to me. Also, when deleting a layer, we need to take care of the parameter calculation, you could discuss this further with @Ram81. I tried to test it in its current state and it doesn't seem to function for a couple of models I imported from the model zoo.

yashdusing:
I thought of the same approach as @Shivanshmundra  did. Isn't it easier to implement a multiple delete button rather than a delete mode button ? What changes are needed to this ?
Also could you tell me about the exact layers it didn't work for

Ram81:
@yashdusing the delete key approach is better from user perspective.

thatbrguy:
@Shivanshmundra @yashdusing If both of you want to pursue it, one of you can work on using the delete key for multiple layers, the other can move the delete button to the actions bar. Sorry for the confusion by calling it "delete mode". We just wanted it to be moved where it's common to all layers (Like the actions bar). 

The code for parameter count update should be available for single layer deletion, be sure to call it wherever necessary.

PR #331, opened 2018-03-15
Fix: Dim Error when adding input layers (#285)

### LINKED ISSUE #285: Nondescript 'dim' error when adding certain data layers
- **Steps to reproduce the problem:** add an ImageData, Data, HD5Data, WindowData or MemoryData layer.

- **The expected behaviour:** Layer is added without incident.
- **The observed behaviour:** Error that says 'dim', and nothing else. [Example.](https://i.imgur.com/fcMSAsG.mp4)

This seems to be an issue with the backend, can I work on this?

Ram81:
Fixes #285.


Ram81:
@utsavgarg The error messages are popping up because if input dimensions of layers are invalid, so let's say the image data layer has a shape of [3, 0, 0] initially and we are stacking a conv3d layer below then issue comes. So I have added error handling code for the issues.
And I have rectified the issues with Data and WindowData layers.

PR #336, opened 2018-03-20
Added variants of ResNeXt to model zoo

yashdusing:
Added variants of ResNext (50, 101, 152) to model zoo.
Paper : https://arxiv.org/abs/1611.05431


yashdusing:
@utsavgarg @thatbrguy @Ram81 could you check this ?

Issue #340, opened 2018-04-21
No support for Linear Activation


MuLx10:
![image](https://user-images.githubusercontent.com/23444642/39081617-6433f4dc-4562-11e8-8c45-1c597f3d04f5.png)

Current keras app does not support linear activation

sd1998:
@MuLx10 Are you still working on this. If not I would like to work on this issue.

sashank27:
@MuLx10 @RishabhJain2018 I would like to work on this, if it is available

Ram81:
@sashank27 sure you can go ahead

PR #341, opened 2018-04-21
Added Linear Activation in Keras app

MuLx10:
Fixes (#340 ) extending support for linear activation function to incorporate models like YOLOv3. Currently works only for keras.

### DIFF
--- a/ide/static/js/data.js
+++ b/ide/static/js/data.js
@@ -2531,6 +2531,36 @@ export default {
     },
     learn: false
   },
+  Linear: {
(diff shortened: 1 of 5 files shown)


thatbrguy:
@MuLx10 Hey, it would be great if you could add the corresponding unittests for the new functionality, in test_views.py

MuLx10:
I'll add it

Ram81:
@MuLx10 are you still working on this?

Ram81:
No activity since 10 sep

PR #342, opened 2018-04-21
Added YOLO v3 and Tiny YOLO

### LINKED ISSUE #340: No support for Linear Activation

MuLx10:
Added keras models for YOLOv3 and TinyYOLO.(#105). Currently export works only for keras. However, it requires Linear Activation layer (Issue #340 ) (Fix #341).


Ram81:
No activities since 21 Apr

PR #344, opened 2018-05-01
Docker deployment

### LINKED ISSUE #130: Setup Docker containers
## Current Scenario

Currently setting up Fabrik on a local machine for development requires running a lot of commands. We want to setup the docker containers so that setting up Fabrik is not pain. 

## Deliverables

A basic setup that fulfills the following requirements:

- [ ] Different containers for Postgres, NodeJS and Django
- [ ] Add instructions in README

Please comment for if you have questions.

Ram81:
Fixes #130.


deshraj:
@Ram81 can you please remove the `docker/prod` for now since that is not perfect. We will add it later when we will work on it.

Ram81:
@deshraj I have added init.py for settings similar to evalai we can use this same when we will be adding support for test and prod.

thatbrguy:
@Ram81 In container-start.sh, keras backend is configured to be theano. Could you change it to TF?

PR #350, opened 2018-05-13
Extend TF Support for Deconv Layer

Ram81:
@deshraj @utsavgarg This PR adds support for Deconvolution Layer in TF Import. It works for TF Import and we are able to export in caffe and keras both. Export for TF works but there is an issue in shapes assignment. I am currently looking into it.

### DIFF
--- a/example/keras/SeparableConvKerasTest.json
+++ b/example/keras/SeparableConvKerasTest.json
@@ -1,0 +1,110 @@
+{
+    "class_name": "Sequential", 
+    "keras_version": "2.0.8", 
+    "config": [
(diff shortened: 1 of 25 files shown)


kdexd:
@Ram81 this requires a rebase now :slightly_smiling_face:

kdexd:
On a related note, repeated merging from master always felt tedious to me. My recommendation is interactive rebase -- see if you find it convenient from now: [blog post](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history)

PR #352, opened 2018-05-17
Add Support for 3D Convolution Layer

Ram81:
@deshraj @utsavgarg This PR adds support for 3D Conv layer + import test.

PR #353, opened 2018-05-20
Extend TF Import Support for Pool layers

Ram81:
@deshraj @utsavgarg In this PR I have extended support for Max Pooling 3D, Average Pooling 3D layer and added tests for the same. This PR depends on #352 , once it is merged there will be some minor changes in tests that need to be made.

### DIFF
--- a/tensorflow_app/views/import_graphdef.py
+++ b/tensorflow_app/views/import_graphdef.py
@@ -10,7 +10,8 @@ from urlparse import urlparse
 
 # map from operation name(tensorflow) to layer name(caffe)
 op_layer_map = {'Placeholder': 'Input', 'Conv2D': 'Convolution', 'Conv3D': 'Convolution',
-                'MaxPool': 'Pooling', 'MatMul': 'InnerProduct', 'Relu': 'ReLU',
+                'MaxPool': 'Pooling', 'MaxPool3D': 'Pooling', 'AvgPool3D': 'Pooling',
+                'MatMul': 'InnerProduct', 'Relu': 'ReLU',
                 'Softmax': 'Softmax', 'LRN': 'LRN', 'Concat': 'Concat',
                 'AvgPool': 'Pooling', 'Reshape': 'Flatten', 'LeakyRelu': 'ReLU',
                 'Elu': 'ELU', 'Softsign': 'Softsign', 'Softplus': 'Softplus'}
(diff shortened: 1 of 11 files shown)


tensorflow_app/views/import_graphdef.py
```diff
@@ -55,17 +56,23 @@ def get_padding(node, layer):
     pad_top = pad_along_height / 2
     pad_left = pad_along_width / 2
     '''
-    if node.type == "Conv3D":
+
+    if node.type == "Conv3D" or node.type == "MaxPool3D" or node.type == "AvgPool3D":
```
kdexd: `node.type in {"Conv3D", "MaxPool3D", "AvgPool3D"}` will make it cleaner and you can extend it without too much or conditions whenever you need to add more support of 3D layers.

tensorflow_app/views/import_graphdef.py
```diff
@@ -244,8 +252,30 @@ def import_graph_def(request):
                         return JsonResponse({'result': 'error', 'error':
                                              'Missing shape info in GraphDef'})
 
+                if str(node.type) == 'MaxPool3D':
```
kdexd: Why do you need to explicitly apply `str` method? I saw it in a few places above -- `node.type` is already a string. I am a bit confused myself.

tensorflow_app/views/import_graphdef.py
```diff
@@ -262,6 +292,28 @@ def import_graph_def(request):
...
+                    layer['params']['stride_w'] = int(
+                        node.get_attr('strides')[3])
+                    layer['params']['layer_type'] = '3D'
+                    try:
+                        layer['params']['pad_d'], layer['params']['pad_h'], \
+                            layer['params']['pad_w'] = get_padding(node, layer)
+                    except TypeError:
+                        return JsonResponse({'result': 'error', 'error':
+                                             'Missing shape info in GraphDef'})
+
```
kdexd: These lines are duplicated in `MaxPool3D`. You can try to make it hierarchical -- first, check if `node.type` has a substring`3D` and then check if it is max or average. That should reduce a lot of duplication.

kdexd: Similar comment goes for `MaxPool` and `AvgPool` variants.

kdexd:
How about aliasing `MaxPool` as `MaxPool2D` and `AvgPool` as `AvgPool2D`, I think Tensorflow layers are named like this too. It would avoid confusion.

Thoughts @Ram81 @deshraj @utsavgarg ?

Ram81:
@kdexd the name's of ops in pbtxt are MaxPool and AvgPool that's why we use it. I am not sure if I get where you want aliasing.

kdexd:
@Ram81 Oh okay, understood. Nevermind.

PR #356, opened 2018-05-28
LRN Support for Keras Import and Export

Ram81:
@utsavgarg this PR adds support for LRN layer in Keras import and export both. Currently I have tested for AlexNet and added test regarding the same. It is built on top of #287.


Ram81:
Sure, alexnet works fine for now. I'll do the required fixes

PR #363, opened 2018-06-19
WIP: Login Support for Fabrik

Ram81:
@deshraj @kdexd @ayshrv @utsavgarg this PR adds support for user login in Fabrik using social apps using django allauth.


pavanagrawal123:
Oh I'm glad the previous code came into use! I remember when I had authored this PR, I had accidentally left a debug page in there in the templates folder, and I see it's still there. If you don't think we need it, maybe it should be removed.

Ram81:
Yeah

backendAPI/templates/login.html
```diff
@@ -0,0 +1,9 @@
+<!-- This is purely for debugging purposes. This will not be served to frontend users. -->
```
utsavgarg: Do we need this file?

backendAPI/tests.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
```
utsavgarg: Do we need this?

backendAPI/templates/login.html
Ram81: No, I''ll remove it

backendAPI/tests.py
Ram81: No, I'll remove this

utsavgarg:
@Ram81 LGTM :) Merging this for now so you are not blocked, but please remove the contents of `backendAPI/models.py` in a future PR.

PR #365, opened 2018-06-26
Update package.json - Remove extraneous package.

pavanagrawal123:
Originally, #363 used the react social login buttons for the icons for homepage, but now it is switched to Font-Awesome, so it is no longer needed


Ram81:
@pavanagrawal123 Nice catch

PR #371, opened 2018-07-29
WIP: Optimize RTC Updates and RTC Highlighting

Ram81:
Work in progress

### DIFF
--- a/backendAPI/views.py
+++ b/backendAPI/views.py
@@ -8,7 +8,7 @@ def check_login(request):
     try:
         user = User.objects.get(username=request.user.username)
         user_id = user.id
-        username = ''
+        username = 'Kafka'
 
         is_authenticated = user.is_authenticated()
         if (is_authenticated):
--- a/caffe_app/consumers.py
+++ b/caffe_app/consumers.py
@@ -76,7 +76,6 @@ def ws_receive(message):
         user_id = data['userId']
         highlight_color = data['highlightColor']
         username = data['username']
-        print(data['username'])
 
         Group('model-{0}'.format(networkId)).send({
             'text': json.dumps({
--- a/caffe_app/consumers.py
+++ b/caffe_app/consumers.py
@@ -74,6 +74,9 @@ def ws_receive(message):
         add_highlight_to = data['addHighlightTo']
         remove_highlight_from = data['removeHighlightFrom']
         user_id = data['userId']
+        highlight_color = data['highlightColor']
+        username = data['username']
+        print(data['username'])
 
         Group('model-{0}'.format(networkId)).send({
             'text': json.dumps({
@@ -81,7 +84,9 @@ def ws_receive(message):
                 'removeHighlightFrom': remove_highlight_from,
                 'userId': user_id,
                 'action': action,
-                'randomId': randomId
+                'randomId': randomId,
+                'highlightColor': highlight_color,
+                'username': username
             })
         })
     else:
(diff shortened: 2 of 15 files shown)


kdexd:
Please resolve the conflicts and ping here :) I'll review it thereafter.

Ram81:
@kdexd Conflicts resolved

caffe_app/consumers.py
```diff
@@ -41,29 +66,109 @@ def ws_receive(message):
...
+    randomId = data['randomId']
+    networkId = message.channel_session['networkId']
+
+    if (action == 'UpdateHighlight'):
+        add_highlight_to = data['addHighlightTo']
+        remove_highlight_from = data['removeHighlightFrom']
+        user_id = data['userId']
+        highlight_color = data['highlightColor']
+        username = data['username']
+        print(data['username'])
```
utsavgarg: Debugging statement?

backendAPI/views.py
```diff
@@ -8,7 +8,7 @@ def check_login(request):
     try:
         user = User.objects.get(username=request.user.username)
         user_id = user.id
-        username = ''
+        username = 'Kafka'
```
utsavgarg: Do we need to keep the default user name as this? Because in the open link sharing case, all users will be given the same user name and it might be confusing. For now can we do something simple like `User1`, `User2` etc. ?

ide/static/js/content.js
```diff
@@ -104,13 +106,25 @@ class Content extends React.Component {
...
     this.handleClick = this.handleClick.bind(this);
     this.performSharedUpdate = this.performSharedUpdate.bind(this);
+    this.performSharedAdd = this.performSharedAdd.bind(this);
+    this.performSharedDelete = this.performSharedDelete.bind(this);
+    this.addHighlightOnLayer = this.addHighlightOnLayer.bind(this);
+    this.addSharedComment = this.addSharedComment.bind(this);
     this.changeCommentOnLayer = this.changeCommentOnLayer.bind(this);
+    this.getRandomColor = this.getRandomColor.bind(this);
+  }
+  getRandomColor() {
```
utsavgarg: Instead of random colours, can we choose from a colour palette to ensure the colours are different? Because in some cases they are very similar. We can have a set of say 10 colours and then reuse them if there are more users.

ide/static/js/content.js
```diff
@@ -129,19 +143,60 @@ class Content extends React.Component {
   onSocketMessage(message) {
     // message received on socket
     let data = JSON.parse(message['data']);
-    let rebuildNet = false;
-    let nextLayerId = this.state.nextLayerId;
+    //let rebuildNet = false;
```
utsavgarg: Are these required?

ide/static/js/layer.js
```diff
@@ -43,6 +43,25 @@ class Layer extends React.Component {
...
+    let highlightUser = null;
+    let highlightClass = '';
+    let highlightColor = '#000';
+
+    if(this.props.layer.highlight && this.props.layer.highlight.length > 0) {
+      highlightClass = 'highlighted';
+      highlightColor = this.props.layer.highlightColor[this.props.layer.highlightColor.length - 1];
+      highlightUser = (<div style={{
+                              background: highlightColor,
+                              color: 'white',
```
utsavgarg: Can we shift these style specifications to the css stylesheet as they are fixed?

caffe_app/consumers.py
Ram81: Sorry for that :sweat_smile:

ide/static/js/content.js
Ram81: Removing them

PR #390, opened 2018-10-23
Support for Real Time Collaboration and Model Export using Celery

Ram81:
@RishabhJain2018


docker/prod/django/supervisor.conf
```diff
@@ -0,0 +1,35 @@
...
+autostart=true
+autorestart=true
+
+redirect_stderr=true
+
+stopasgroup = true
+
+[supervisorctl]
+
+[supervisord]
```
RishabhJain2018: Please add a new line here.

PR #393, opened 2018-10-25
Graphs are reset after import and export for Keras and Tensorflow

Azamlynny:
Resets graph after export and before import for Keras and TensorFlow

### DIFF
--- a/ide/static/js/content.js
+++ b/ide/static/js/content.js
@@ -1103,19 +1103,32 @@ class Content extends React.Component {
   faqModal() {
     this.modalHeader = "Help/FAQ"
     this.modalContent = (<p><b>Q:</b> What is Fabrik?<br />
-      <b>A:</b> Fabrik is an online platform, created by CloudCV, allowing AI researchers and enthusiasts to build and visualize deep learning models.<br />
+      <b>A:</b> Fabrik is an online platform, created by CloudCV, allowing AI researchers and enthusiasts to
+      build and visualize deep learning models.<br />
       <b>Q:</b> What is the model zoo?<br />
-      <b>A:</b> It is a collection of pre-built models that you can use. To access it, simply click the folder icon in the left corner of the toolbox and pick a model. You can find the available models <a target="_blank" href="https://github.com/Cloud-CV/Fabrik/blob/master/docs/source/tested_models.md">here</a>.<br />
+      <b>A:</b> It is a collection of pre-built models that you can use.
+      To access it, simply click the folder icon in the left corner of the toolbox and pick a model.
+      You can find the available models
+      <a target="_blank" href="https://github.com/Cloud-CV/Fabrik/blob/master/docs/source/tested_models.md"> here</a>.
+      <br />
       <b>Q:</b> What do the Train/Test buttons mean?<br />
-      <b>A:</b> They are two different modes of your model: Train and Test - respectively for training your model with data and testing how and if it works.<br />
+      <b>A:</b> They are two different modes of your model: 
+      Train and Test - respectively for training your model with data and testing how and if it works.<br />
       <b>Q:</b> What does the import fuction do?<br />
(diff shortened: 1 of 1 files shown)


tensorflow_app/views/export_graphdef.py
```diff
@@ -1,20 +1,20 @@
 import os
 import string
 import random
+import tensorflow as tf
 from django.views.decorators.csrf import csrf_exempt
 from django.http import JsonResponse
 from keras_app.views.export_json import export_json
+from tensorflow.keras.backend import clear_session
```
thatbrguy:
```suggestion
from keras.backend import clear_session
```
Looks like travis build is failing when we try to access `tensorflow.keras.backend`. Can you try this instead? If this suggestion also fails travis build, we can try using `from keras import backend as K` and then call `K.clear_session()` instead. Try making this change at every instance where you used `keras.backend`.

tensorflow_app/views/export_graphdef.py
```diff
@@ -26,6 +26,10 @@ def export_to_tensorflow(request):
     os.chdir(BASE_DIR + '/tensorflow_app/views/')
     os.system('KERAS_BACKEND=tensorflow python json2pbtxt.py -input_file ' +
               randomId + '.json -output_file ' + randomId)
+    
+    tf.keras.backend.clear_session()
```
thatbrguy: Same issue as above, try using `keras.backend` instead. And since you have imported `clear_session()`, why not just directly call `clear_session()`? (Instead of `tf.keras.backend.clear_session()`).

tensorflow_app/views/import_graphdef.py
```diff
@@ -8,6 +8,7 @@
 import re
 import urllib2
 from urlparse import urlparse
+from tensorflow.keras.backend import clear_session
```
thatbrguy: Same issue as above, try using `keras.backend` instead.

tensorflow_app/views/import_graphdef.py
```diff
@@ -149,7 +150,8 @@ def import_graph_def(request):
                 return JsonResponse({'result': 'error', 'error': 'Invalid URL\n'+str(ex)})
         else:
             return JsonResponse({'result': 'error', 'error': 'No GraphDef model found'})
-
+        
+        tf.keras.backend.clear_session()
```
thatbrguy: Same issue as above, try using keras.backend instead. And since you have imported clear_session(), why not just directly call clear_session()? (Instead of tf.keras.backend.clear_session()).

keras_app/views/export_json.py
```diff
@@ -26,6 +27,9 @@ def randomword(length):
 
 @csrf_exempt
 def export_json(request, is_tf=False):
+    clear_session
```
thatbrguy:
```suggestion
    clear_session()
```
Function call

keras_app/views/import_json.py
```diff
@@ -14,10 +14,13 @@
 from keras.models import model_from_json, Sequential
 from keras.layers import deserialize
 from ..custom_layers.lrn import LRN
-
+from keras.backend import clear_session
 
 @csrf_exempt
 def import_json(request):
+    clear_session
```
thatbrguy:
```suggestion
    clear_session()
```
Function call

thatbrguy:
@Ram81 @Azamlynny Looks like using `keras.backend` instead of `tf.keras.backend` fixed the travis issue. However, flake8 failed so please see to that. Just run the command `flake8` in the Fabrik repository and correct the errors.

Otherwise, I've run it on my local machine and everything seems fine (after correcting the lines I marked in my review).

thatbrguy:
@Ram81 LGTM :+1:

thatbrguy:
@Azamlynny Hey, could you add the changes relevant to your new task in a separate PR?

PR #412, opened 2018-10-30
Add login page

Jatana:


RishabhJain2018:
@Jatana Can you please resolve the conflicts?

Ram81:
@Jatana the work looks pretty good for initial work, we would definitely need to iterate over the design of login page. Please have a look at https://evalai.cloudcv.org/ login page. It would be good if you can come up with some mock-ups for login page design so that we can iterate on this further.

Ram81:
Redundant to #424

PR #426, opened 2018-11-04
Fix #340: Added linear activation support for Keras

### LINKED ISSUE #340: No support for Linear Activation

haabhi:
This PR adds support for both importing and exporting linear activation layer from and for Keras. It also includes test cases for testing both import and export of the layer.


thatbrguy:
@Ram81 @haabhi Also merge the latest changes from the `master` branch. LGTM otherwise.

PR #433, opened 2018-11-07
Fix style issues with login panel

Ram81:


Ram81:
Closing this as it is similar to #435

PR #436, opened 2018-11-07
GCI: Fix the layer choppiness (lag) when a layer is dragged.

ZeroZeroJedenJeden:
Fixed layer lags.

### DIFF
--- a/ide/static/js/canvas.js
+++ b/ide/static/js/canvas.js
@@ -16,12 +16,14 @@ class Canvas extends React.Component {
     this.clickCanvas = this.clickCanvas.bind(this);
     this.clickLayerEvent = this.clickLayerEvent.bind(this);
     this.hoverLayerEvent = this.hoverLayerEvent.bind(this);
+    this.mouseUpEvent = this.mouseUpEvent.bind(this);
     // whether a layer was clicked or dragged
     this.clickOrDraggedLayer = 0;
     this.hover = 0;
     this.mouseState = null;
     this.placeholder = true;
     this.disableZoom = true;
+    this.lastEvent = null;
   }
   /* this function returns the layers between a specified output y and input y
   it also sneaks in another functionallity of determining which direction is most crowded. this is specifically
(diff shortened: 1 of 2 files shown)


yashdusing:
Pretty good @ZeroZeroJedenJeden . The layers seem responsive now. Will see if there's any bugs but so far : LGTM

ide/static/js/canvas.js
```diff
@@ -272,13 +274,18 @@ class Canvas extends React.Component {
...
-    let net = this.props.net
-    this.checkCutting(net);
+    this.lastEvent = event;
+  }
+  mouseUpEvent(event, layerId){
+    if(this.lastEvent) {
+      const layer = this.props.net[layerId];
+      layer.state.left = `${this.lastEvent.pos['0']}px`;
+      layer.state.top = `${this.lastEvent.pos['1']}px`;
+      this.props.modifyLayer(layer, layerId);
```
Ram81: @ZeroZeroJedenJeden rather than modifying the layer details for every mouse event can't we have a regular interval let's say 5 events and then make update. Because the current implementation still doesn't help much with the issue.

Ram81: Because modifyLayer call does quite a lot of computation which is not needed to be changed for every mouse event.

ZeroZeroJedenJeden: mouseUpEvent activate when you release left mouse buttom.

ZeroZeroJedenJeden: So, "mouseUpEvent" function is executed only one time for all dragging one element.

Ram81:
@ZeroZeroJedenJeden work looks good on the end of layer choppiness but can you figure out an alternative to calling modifyLayer on every mouse event because it does many more operations that changing the state of layer object like parameter updation, total parameter calculation, etc.

To update top and left co-ordinates of layer object you can do it explicitly rather than using modifyLayer. It is quite inefficient implementation right now. 
Please let me know if I can help with any doubts you have to fix this.

ide/static/js/canvas.js
Ram81: Okay, let me test on my system.

yashdusing:
Before the PR : 
ezgif-5-635b99468a14
After the PR :
ezgif-5-5d8b0a0fe5b7

PR #443, opened 2018-11-10
Fixed the padding calculation bug

sotanishy:
When the padding is not an integer, like 1/2 or -1/2, it should be rounded up but instead it was rounded down, causing errors.


Ram81:
@sotanishy GoogleNet export still doesn't work can you look into that.

sotanishy:
Xception didn't work before but after the change it worked.

Ram81:
@sotanishy can you test whether all the models mentioned in README are working fine or is there some model which is breaking.

sotanishy:
I haven't tested everything yet but GoogleNet export for keras worked.

Ram81:
@sotanishy can you fix the Travis CI issues and test all the models listed in tested models (check if any other model which was working earlier is not breaking after the change). Once you are done with that we can go forward with the review.

sotanishy:
Importing keras models such as inception and VQA failed. I'm looking into why but I doubt it's related to the changes I made. Import and export for other models currently available in keras worked. GoogleNet and DenseNet, which were not available in keras before, were successfully exported to keras, too.

Ram81:
@sotanishy the inception models were working fine before the change right?
I encountered a similar issue while fixing the bug can you please verify once again and if needed please fix the issue.

sotanishy:
Exporting and importing them again in caffe worked, but the parameter count was more than what's mentioned in the issue.

PR #450, opened 2018-11-14
Updated for protection against MIME & XSS based attacks

yashdusing:
Updated headers to ajax call for protection against MIME and XSS based attacks


yashdusing:
It shows up as unprotected (the same as shown in #450). Although I do have a doubt. 0.0.0.0:8000 is the home page site which has no headers on it. The headers are added to the ajax calls made to import/export or other apps so isn’t it supposed to show up not protected ?(unless we somehow added XSS protection to home page)

yashdusing:
So 0.0.0.0:8000/layer_parameter also is shown as unprotected 😓. I will have to figure out why

yashdusing:
screenshot_20181117_235839
Updated for main page with postman results

PR #455, opened 2018-11-15
Fix #447: Changed format for exporting models to MetaGraphDef for Tensorflow

### LINKED ISSUE #447: Models exported as graphdef protos for Tensorflow cannot be trained.
When graphdefs are imported using tf.import_graph_def(), variable collections within tensorflow such as GLOBAL_VARIABLES and TRAINABLE_VARIABLES are not updated. Since tensorflow optimizers look for variables to optimize in the above collections by default, calling tf.train.Optimizer.minimize() fails with an error saying that there are no variables to be optimized.

This strangely seems to be an issue only with graphdefs extracted from the keras backend session and not a problem with Fabrik. 

A better way would be to export metagraphdefs, which contain the graphdef along with information about variable collections, which are restored when the metagraphdef is imported. Exporting models for tensorflow as .meta files can be done with very little change in the export logic as it is now.

haabhi:
This PR makes Fabrik export models as MetaGraphDef instead of Graphdef, which fixes #447.  Doing this allows models built in Fabrik to be imported into Tensorflow seamlessly.

### DIFF
--- a/ide/tasks.py
+++ b/ide/tasks.py
@@ -311,19 +311,22 @@ def export_keras_json(net, net_name, is_tf, reply_channel):
         json_str = json_str.strip("'<>() ").replace('\'', '\"')
         lrnLayer = imp.load_source('LRN', BASE_DIR + '/keras_app/custom_layers/lrn.py')
 
+        # clear clutter from previous graph built by keras to avoid duplicates
+        K.clear_session()
+
         model = model_from_json(json_str, {'LRN': lrnLayer.LRN})
 
-        sess = K.get_session()
-        tf.train.write_graph(sess.graph.as_graph_def(add_shapes=True), output_fld,
-                             output_file + '.pbtxt', as_text=True)
+        tf.train.export_meta_graph(
+            os.path.join(output_fld, output_file + '.meta'),
+            as_text=True)
 
         Channel(reply_channel).send({
             'text': json.dumps({
                 'result': 'success',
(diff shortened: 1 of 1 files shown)


yashdusing:
@haabhi are you going to make the changes in import_graphdef as well ?

haabhi:
@yashdusing Yes, I am doing that at the moment. I am working on importing both metagraphdefs and graphdefs so the model zoo models and models exported from fabrik both work.

ide/tasks.py
```diff
@@ -311,19 +311,22 @@ def isProcessPossible(layerId):
...
 
+        # clear clutter from previous graph built by keras to avoid duplicates
+        K.clear_session()
+
         model = model_from_json(json_str, {'LRN': lrnLayer.LRN})
 
-        sess = K.get_session()
-        tf.train.write_graph(sess.graph.as_graph_def(add_shapes=True), output_fld,
-                             output_file + '.pbtxt', as_text=True)
+        tf.train.export_meta_graph(
```
Ram81: @haabhi can you make the same changes in `tensorflow_app/views/export_grahdef.py`. As of now there are no test cases for celery tasks which is why we run tests on our older implementation and it's necessary to keep them synced.

haabhi: Alright, I am on it :)

haabhi: @Ram81 Done :+1:

PR #470, opened 2018-11-22
Add a new test case for Keras shape calculation

sotanishy:
This is the corner case I came across when I tried exporting Xception model. The test fails as expected.


sotanishy:
@Ram81 #443 solves the issue.

sotanishy:
Merged the changes in #443 into this branch.

PR #471, opened 2018-11-23
Dashboard (Both design and functionality)

sayamkanwar:
Hey,
I had claimed the task '[Fabrik] Add backend support for maintaining dashboard for a user' but there was no dashboard in Fabrik presently so I had to design a dashboard first which was another task (https://codein.withgoogle.com/tasks/6415117748011008/). So I have made the deliverables for both the tasks because the former task wasn't possible without completing the latter task. 

# Work done
1. I have created a new save button in the canvas sidebar (under Actions), I have created a function `saveModel_to_db` for that, it saves the model to the database with `public_sharing` set to false by default. This function first checks the database if the model already exists, it updates the JSON field instead of creating a new entry if the model already exists.

screenshot 2018-11-24 at 2 13 40 am

On clicking the save button, if there are no errors. It gives the user a success message.

screenshot 2018-11-24 at 2 16 44 am

2. I have added a button 'Dashboard' in canvas sidebar which takes the user to dashboard. This button is only shown when the user is logged in, also the if someone tries to open dashboard by manually entering the dashboard URL, it redirects the person to home page (i.e. canvas) if the user is not logged in.

3. The dashboard consists of a button 'Create a new model' and it displays the list of models created by the user with it's name and thumbnail. Each model list item has two options: edit and delete which become visible on hover. Currently the thumbnail is just a sample image as I didn't know how will the actual preview of the model will be rendered as a thumbnail and I was guided by the mentor to just leave it like that for the time being. I have also created the route to dashboard(/dashboard).

screenshot 2018-11-24 at 2 27 44 am

screenshot 2018-11-24 at 2 27 58 am
[on hover]

screenshot 2018-11-24 at 2 30 41 am
[on deleting the model]

screenshot 2018-11-24 at 2 31 23 am
[if the user hasn't created any models]

4. I have modified the existing `load_from_db` function, now it also returns the `public_sharing` property of the model which is used by the main canvas to check if a model is shared. If it's shared it adds comments option to it and other shared features. If it's not shared, it acts as an editor for the user.

5. The current criteria for rendering the Share/History button in the canvas was that if the path is /load so I had to update the criteria, now I'm passing on the value of `public_sharing` to the `topBar` component as a prop value and in the `topBar` component it renders the Share button if value of `isShared` state is false, else if it's true it renders the History button which shows the history of the model.

6. To make login system more easy to work with, I have added the piece of code in the login function to set localStorage item 'userID' to the value of the user id, it makes the work a lot more easier as different components can use user id value for different functions. On logout, the localStorage item is destroyed as well.
NOTE: I haven't removed any of the already existing code for the login part, I have just made an addition.

7. I have added the tests for all the new functions that I have created in `tests/unit/caffe_app/test_db.py`.

screenshot 2018-11-24 at 2 43 17 am
[all tests passed]

# Modifications in old features
In the `save_to_db` function, now it checks if the model exists in the database and updates the json field else it creates a new record in the database. Also, if a model exists in the database with public_sharing set to false. On clicking the share button which triggers the `save_to_db` function sets the public_sharing of the existing model to true instead of creating a new shared model.

# Correction
For the people who are using docker to run Fabrik, they'll have to edit the `settings/test.py` file because the variables have been set to the values being used in the virtual environment method, I have added the comments in this file for the changes that have to be made for docker, otherwise it will throw an error.

Please review my work @Ram81 @RishabhJain2018, thank you! :)

### DIFF
--- a/ide/static/js/dashbutton.js
+++ b/ide/static/js/dashbutton.js
@@ -11,7 +11,7 @@ class DashButton extends React.Component {
 	render(){
 		return(
 			<div>
-			<h5 style={{marginBottom: '0.6em'}} className="sidebar-heading" id="sidebar-login-button" onClick={() => {this.openDash()}}>
+			<h5 className="sidebar-heading" id="sidebar-dash-button" onClick={() => {this.openDash()}}>
 			<div>DASHBOARD</div>
 			</h5>
 			</div>


ide/static/js/content.js
```diff
@@ -979,6 +980,35 @@ class Content extends React.Component {
     layer.info.phase = 0;
     this.setState({ net });
   }
+  saveModel(){
+  let netData1 = this.state.net;
```
Ram81: Please rename this variable

ide/static/js/content.js
```diff
@@ -979,6 +980,35 @@ class Content extends React.Component {
...
+  console.log(netData1);
```
Ram81: remove log statements

ide/static/js/content.js
```diff
@@ -1055,7 +1086,7 @@ class Content extends React.Component {
     // Note: this needs to be improved when handling conflict resolution to avoid
     // inconsistent states of model
     let nextLayerId = this.state.nextLayerId;
-
+    var shareBool = false;
```
Ram81: rename the variable and use `let`

ide/static/js/content.js
```diff
@@ -1299,9 +1341,11 @@ class Content extends React.Component {
           <div id="sidebar-scroll" className="col-md-12">
              <h5 className="sidebar-heading">ACTIONS</h5>
              <TopBar
+              isPublic_Sharing={this.state.isShared}
```
Ram81: Remove `_`

ide/static/js/dashboard.js
```diff
@@ -0,0 +1,238 @@
...
+    marginRight           : '-50%',
+    transform             : 'translate(-50%, -50%)',
+    borderRadius          : '8px'
+  },
+  overlay: {
+    zIndex                : 100
+  }
+};
+
+class Dashboard extends React.Component{
```
Ram81: Don't add two different components  in same files

ide/urls.py
```diff
@@ -1,9 +1,10 @@
+# flake8: noqa
 from django.conf.urls import url, include
 from django.contrib import admin
 from django.conf.urls.static import static
 from django.conf import settings
 from views import index, calculate_parameter, fetch_layer_shape
-from views import load_from_db, save_to_db, fetch_model_history
+from views import load_from_db, loadModel_from_db, deleteModel_from_db, save_to_db, saveModel_to_db, fetch_model_history
```
Ram81: please follow naming conventions for method names

ide/views.py
```diff
@@ -93,29 +166,59 @@ def save_to_db(request):
...
-            model_update = NetworkUpdates(network_version=model_version,
-                                          updated_data=json.dumps({'nextLayerId': next_layer_id}),
-                                          tag='ModelShared')
-            model_update.save()
-
-            return JsonResponse({'result': 'success', 'id': model.id})
-        except:
-            return JsonResponse({'result': 'error', 'error': str(sys.exc_info()[1])})
 
+        if Network.objects.filter(name=net_name).exists():
```
Ram81: net_nam is not the primary key, please use mode id as the key to filter

ide/views.py
```diff
@@ -216,20 +319,37 @@ def load_from_db(request):
...
+                                 'next_layer_id': next_layer_id, 'public_sharing': model.public_sharing})
 
     if request.method == 'GET':
         return index(request)
 
 
+@csrf_exempt
+def loadModel_from_db(request):
+    if request.method == 'POST':
+        if 'userID' in request.POST:
```
Ram81: Please follow naming conventions

ide/views.py
```diff
@@ -82,6 +83,78 @@ def calculate_parameter(request):
...
+            model = Network.objects.get(id=model_id)
+            if model.author_id == int(userID):
+                model.delete()
+                return JsonResponse({'result': 'success', 'data': 'Model successfully deleted!'})
+            else:
+                return JsonResponse({'result': 'error', 'error': "This model doesn't belong to you!"})
+
+
+@csrf_exempt
+def saveModel_to_db(request):
```
Ram81: Why do we need to save model to db methods?

sayamkanwar: How will I check using model id? If a person is creating a new model with the same model name then only I can run a query right?

ide/views.py
sayamkanwar: How will I check using model id? If a person is creating a new model with the same model name then only I can run a query right?

ide/views.py
sayamkanwar: The save button triggers this function

ide/views.py
sayamkanwar: Basically, in this method I'm taking the net name from the POST request to filter from database because that's the only parameter that can be used to do this as this would be the only similarity amongst the already existing records and new models. So I can't really use model id to filter out results.

sayamkanwar:
@Ram81, I have made all the changes except 2, I have replied so as to why I can't make those changes, please read those and let me know what should be done.
Thank you!

ide/static/js/content.js
```diff
@@ -979,6 +980,34 @@ class Content extends React.Component {
     layer.info.phase = 0;
     this.setState({ net });
   }
+  saveModel(){
```
Ram81: As far as I understand this method performs update on already existing model so please rename it to update model

ide/static/js/content.js
```diff
@@ -1072,6 +1101,13 @@ class Content extends React.Component {
           // while loading a model ensure paramete intialisation
           // for UI show/hide is not executed, it leads to inconsistent
           // data which cannot be used further
+          if (response.public_sharing == false) {
+            is_shared = false;
```
Ram81: If you are not changing value before this step then why do you have to set it to false again as the default value is already false

ide/static/js/content.js
```diff
@@ -1299,9 +1339,11 @@ class Content extends React.Component {
           <div id="sidebar-scroll" className="col-md-12">
              <h5 className="sidebar-heading">ACTIONS</h5>
              <TopBar
+              isPublicSharing={this.state.isShared}
               exportNet={this.exportNet}
               importNet={this.importNet}
               saveDb={this.saveDb}
+              saveModel={this.saveModel}
```
Ram81: Change the method name to update model (if I understand the functionality correctly

ide/static/js/content.js
```diff
@@ -1299,9 +1339,11 @@ class Content extends React.Component {
           <div id="sidebar-scroll" className="col-md-12">
              <h5 className="sidebar-heading">ACTIONS</h5>
              <TopBar
+              isPublicSharing={this.state.isShared}
```
Ram81: There is one flag variable in state of content.js  where I maintain the type of model as isShared or not, can't we use that here because defining a new one causes necessity to maintain consistency in both. If not possible then please rename this variable name in props

ide/templates/index.html
```diff
@@ -26,7 +26,7 @@
   </script>
 
 </head>
-<body>
+<body style="background: #F3F5F7;">
```
Ram81: don't add style tags here maintain them in css

ide/urls.py
```diff
@@ -14,9 +17,15 @@
...
     url(r'^model_history', fetch_model_history, name='model-history'),
-    url(r'^model_parameter/', calculate_parameter, name='calculate-parameter'),
-    url(r'^layer_parameter/', fetch_layer_shape, name='fetch-layer-shape')
-] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT) + \
-    static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
+    url(r'^model_parameter/', calculate_parameter,
+        name='calculate-parameter'),
+    url(r'^layer_parameter/', fetch_layer_shape,
+        name='fetch-layer-shape'),
+    ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT) \
```
Ram81: Fix alignment

ide/static/js/content.js
sayamkanwar: This method doesn’t update on already existing model, it saves the model data on clicking the save button but if the model already exists in the database then it updates the model.

ide/static/js/content.js
sayamkanwar: I haven’t defined this, this was already in the existing code. isShared is the flag variable which is there in content.js file right?

ide/static/js/content.js
sayamkanwar: I have explained this function above, please have a look at it, it’s not just updating the already existing model, it updates the model only if it already exists else it saves the model.

ide/templates/index.html
sayamkanwar: Oh okay!

ide/urls.py
sayamkanwar: This was according to the pep8. It gives an error otherwise.

ide/static/js/content.js
sayamkanwar: Okay, I’ll make the change, thanks for pointing out!

sayamkanwar:
@Ram81, what exactly do you mean by testing it for real time collaboration feature? I have already mentioned the load page is working fine for shared models. If that’s what you are asking?

ide/static/js/content.js
sayamkanwar: I didn’t exactly understand, could you please elaborate?

sayamkanwar:
@Ram81, please have a look at the replies and let me know what shall be done further.
Thank you! :)

Ram81:
@sayamkanwar I meant did you try making updates in real time collaboration mode and see the changes getting reflected in corresponding tabs.

ide/static/js/content.js
Ram81: There is one boolean flag in content.js state variable if you are passing it down as props use the same name and don't define new ones if possible

ide/static/js/content.js
Ram81: So if you are doing this then why do we have older method still present?
 Why can't your remove that or add the code in same method?

sayamkanwar:
Sure, I’ll make the changes!
Thank you

ide/static/js/content.js
sayamkanwar: I can’t remover the older method as there is a difference between the 2 methods the older one is saving the model with shared properties while the new method is saving the model with not shared properties, if you still want me to combine the two, I’ll try to do it.

Ram81:
@sayamkanwar please reopen the PR on `user-dashboard` branch, we'll have to test this thoroughly and release it when login, signup and homepage are ready

PR #476, opened 2018-11-26
GCI: Research RNN implementation

seventeen-iterator:
Made some investigations on implementing recurrence reading for TF import.


gci_rnn_implementation_research/requirements.txt
```diff
@@ -0,0 +1,2 @@
+tensorflow==1.4.1
```
Ram81: Please remove this file

gci_rnn_implementation_research/.gitignore
```diff
@@ -0,0 +1 @@
+*.pyc
```
Ram81: remove this file

seventeen-iterator:
@Ram81 
> But the name patterns are not the best way to do it.
You're right. But if we talk in terms of feasibility and robustness.
> can you share the references you used for the method you specified here
I wasn't sure what type of references you're looking for, because I was developing the algorithm on myself, but I added links to frameworks in the references section.

Ram81:
@seventeen-iterator parsing dependent on name of node defs is not the best way to implement it, you can see my work in the PR I mentioned. There are many cases for which the parsing logic fails. For example if you use the logic for importing a model written in TF then it works but if you export model from Fabrik and try importing it doesn't work because of difference in node defs.

Although the goal of this task was to research for possible methods and you have done good work in doing the same. Your deliverables are good for the task but I would suggest you to spend some more time and see if there is some better way than this. I'll extend your deadline for 2 days. Try to look for some other possible ways.

yashdusing:
Hey ! I haven't looked into RNN representations in tensorflow but I've a few doubts.
So, the problem is RNN is not supported (in Fabrik) because there is no way to identify the basic ops from the pbtxt files ? As mentioned by @haabhi , converting it to basic format like onnx could help. ONNX has experimental support (https://github.com/onnx/tutorials) so another alternative we could look at is mmdnntools. Also, if Keras' RNN support is implemented in Fabrik, we could use MMDNNtools to convert tensorflow pbtxt files directly to json using mmdnntools/onnx and use the json instead. Moreover, we could use this approach for other frameworks as well and extending support would be easy

seventeen-iterator:
@yashdusing Hi! I had the similar idea to export through ONNX to model format with recurrence support for Fabrik import. I even managed to write a program to do so, but I'm facing KeyError. Still don't know the reason of it even after reading tf-onnx issues.
MMdnn tools have [even more experimental support of tensorflow](https://github.com/Microsoft/MMdnn/blob/master/mmdnn/conversion/tensorflow/README.md). Moreover, "currently no RNN related operations support."
@Ram81 now I see, you want not to find other method than parsing nodes, but to remove disadvantage of name dependency. But the only way is to lookup all name spaces to find these that do recurrent job. Then I'll experiment with pure TF, without Keras.

PR #478, opened 2018-11-28
Create model schematic from Fabrik

rpalakkal:
here is the procedure to fully implement this feature along with the code to parse the internal json representation of the model


rpalakkal:
just fixed it

rpalakkal:
@Ram81  Sorry I'm new to this; Wasn't really sure what flake8 is so I just looked up how to get rid of errors(I didn't think that the spacing issues actually made a difference). It should be fixed now:) Thanks

rpalakkal:
I've added 3 sample models and here are the screenshots of the output dimensions of those models. I've also updated the instructions.md, saying to make getDimensions a function and then to call it from the import_json.py,etc. file so that there is no need to save the internal model as a json file.
screen shot 2018-11-29 at 6 40 58 pm
screen shot 2018-11-29 at 6 41 45 pm
screen shot 2018-11-29 at 6 40 25 pm

PalashTanejaPro:
Hey, I tried testing your script with NN-SVG, but the problem I see right now is that it even though the general architecture variable might be mapping, it doesn't work with NN-SVG in practice. 

Here's the testing method I used:
1. Get the value of the architecture2 variable from your code. 
2. Set a breakpoint at the redraw function 
3. Change the value of architecture2 to the one that I got in [1] before the draw is called

screenshot 2018-11-30 at 3 15 53 pm

rpalakkal:
Hi @PalashTanejaPro my bad, I forgot to include the last bit of the getDimensions, and had to tweak another bit of the finding. The format of the architecture variable doesn't have any subarrays so it would be [256,64,64,32,8,8] etc. instead of [[256, 64, 64], [32, 8, 8]. 
image

Also, in your screenshot, no need to change the architecture 2 variable(that is assigned a value based on the architecture variable).

PalashTanejaPro:
@Ram81 @rpalakkal 
That fixes the problems on my end. LGTM 👍 for this task.  
More work will need to be done for correct implementation though, but this is a good guideline to follow.

Ram81:
@PalashTanejaPro sounds good

PR #494, opened 2018-12-08
Added Colornet model

kayo11:
Added Colornet model to the modelZoo(https://github.com/richzhang/colorization/blob/master/models/alexnet_deploy_lab.prototxt):
colornetscr1

colornetscr2


kayo11:
@Ram81 Sure, right on it.

kayo11:
Done: export works for Caffe and I added it to tested_models.md, along with the soource and visualization link.

kayo11:
@Ram81 What do You think?

kayo11:
@Ram81 I double checked,  cloned the repository again and it should be there (at least I see it):

scr

I think it should be there but I'll check once more.

kayo11:
@Ram81 Yes, I'm pretty sure everything is on place.

PR #496, opened 2018-12-09
Share Model URL fix

sayamkanwar:
Hi,
I have fixed the bug in the share model feature which generated different URLs every time. Now it generates only one URL for a model.

Here's a demo

share_patch

Please review @Ram81 @RishabhJain2018 
Thank you! :)

### DIFF
--- a/ide/views.py
+++ b/ide/views.py
@@ -107,14 +107,9 @@ def save_to_db(request):
                 existing_model = NetworkVersion.objects.get(network_id=model_id)
                 existing_model.network_def = net
                 existing_model.save()
-                # create initial update for nextLayerId
-                model_update = NetworkUpdates(network_version=existing_model,
-                                              updated_data=json.dumps({'nextLayerId': next_layer_id}),
-                                              tag="ModelShared")
-                model_update.save()
-                return JsonResponse({'result': 'success','id': model.id})
+                return JsonResponse({'result': 'success', 'id': model.id})
             except:
-                return JsonResponse({'result': 'error','error': str(sys.exc_info()[1])})
+                return JsonResponse({'result': 'error', 'error': str(sys.exc_info()[1])})
         else:
             try:
                 if user_id:
--- a/ide/views.py
+++ b/ide/views.py
@@ -89,32 +89,56 @@ def save_to_db(request):
         net_name = request.POST.get('net_name')
         user_id = request.POST.get('user_id')
         next_layer_id = request.POST.get('nextLayerId')
-        public_sharing = True
         user = None
+        public_sharing = True
         if net_name == '':
             net_name = 'Net'
-        try:
-            # making model sharing public by default for now
-            # TODO: Prvilege on Sharing
-            if user_id:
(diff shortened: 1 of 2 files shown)


sayamkanwar:
Also @Ram81, what else does the task demand? Wasn’t that the only correction in this task?

ide/views.py
```diff
@@ -89,32 +89,51 @@ def save_to_db(request):
         net_name = request.POST.get('net_name')
         user_id = request.POST.get('user_id')
         next_layer_id = request.POST.get('nextLayerId')
-        public_sharing = True
         user = None
+        public_sharing = True
```
Ram81: Why was this necessary?

ide/views.py
```diff
@@ -89,32 +89,51 @@ def save_to_db(request):
...
-            model_update = NetworkUpdates(network_version=model_version,
-                                          updated_data=json.dumps({'nextLayerId': next_layer_id}),
-                                          tag='ModelShared')
-            model_update.save()
-
-            return JsonResponse({'result': 'success', 'id': model.id})
-        except:
-            return JsonResponse({'result': 'error', 'error': str(sys.exc_info()[1])})
+
+        if Network.objects.filter(name=net_name).exists():
```
Ram81: What if filter returns `null`

ide/views.py
```diff
@@ -89,32 +89,51 @@ def save_to_db(request):
...
+            # Update the exising json field
+            try:
+                if user_id:
+                    user_id = int(user_id)
+                    user = User.objects.get(id=user_id)
+                # load the model with the net name
+                model = Network.objects.get(name=net_name)
```
Ram81: Why are we using `net_name` as a unique key?

ide/views.py
```diff
@@ -89,32 +89,51 @@ def save_to_db(request):
...
+                model_id = model.id
```
Ram81: Isn't there any better way to handle this for anonymous sessions?

sayamkanwar:
@Ram81, I'm sorry I had to abandon this task as I could not understand what to do in this. But I would like to work on this after the contest too.

Ram81:
Closing this as student has abandoned the task

ide/views.py
sayamkanwar: We are using ``net_name`` as a unique key because the only similarity between the new model and already existing model would be the model name. That's why I can't use anything else like id.

ide/views.py
sayamkanwar: Answered above

ide/views.py
sayamkanwar: That's why there is an if statement, if this returns true then this will be executed. Else it will create a new model record.

sayamkanwar: If the result is null, it will return false

ide/views.py
sayamkanwar: This was necessary because then the model would not be shared and hence the canvas will not render the comments and sharing stuff on the model.

PR #504, opened 2018-12-12
Fix #391: Added crop_size parameter to prototxt files

### LINKED ISSUE #391: Fix issue with AvgPooling layer padding calculation
Goal of the issue is to fix the bug in padding calculation logic for pooling layers. Due to miscalculation of padding shapes of layers are miscalculated leading to failure in model exports and wrong parameter calculation.

namish800:
Modified prototxt files of the following models:
* MNIST-Lenet
* MNIST-Siamese
* All-CNN

### DIFF
--- a/example/caffe/All_CNN.prototxt
+++ b/example/caffe/All_CNN.prototxt
@@ -7,6 +7,9 @@ layer {
   include {
     phase: TRAIN
   }
+  transform_param{
+    crop_size: 32
+  }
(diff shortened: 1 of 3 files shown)


Ram81:
@namish800 did you test whether import and export works for the new prototxts ?

namish800:
@Ram81 yes import and export is working for new prototxts.

PR #507, opened 2018-12-19
Fix #484: Fixed misalignment issue under 'Wrapper' section

### LINKED ISSUE #484: The items under wrapper are not properly aligned in chrome
The items under wrapper are not properly aligned on chrome.
They are expected to be aligned as under Normalized:

screenshot_bug

I'd like to fix this issue

rahulsnkr:
Fixed the issue of items under the 'Wrapper' section being misaligned (issue #484).

screenshot from 2018-12-19 15-35-51

### DIFF
--- a/ide/static/css/zoo.css
+++ b/ide/static/css/zoo.css
@@ -2,7 +2,7 @@ body {
     overflow-x: hidden;
  }
 
-#wrapper {
+#content-wrapper {
     padding-left: 250px;
 }
 
--- a/ide/static/js/modelZoo.js
+++ b/ide/static/js/modelZoo.js
@@ -98,7 +98,7 @@ class ModelZoo extends React.Component {
 
     return (
       <div className="sidebar-content">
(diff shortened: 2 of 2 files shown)


ide/static/css/zoo.css
```diff
@@ -2,7 +2,7 @@ body {
     overflow-x: hidden;
  }
 
-#wrapper {
+#content-wrapper {
```
Ram81, line 5: @rahulsnkr didn't get this, why would changing the id of the css class worked for you?

rahulsnkr: @Ram81 
https://github.com/Cloud-CV/Fabrik/blob/0bb88a50f570bf443339d4d4f88c66241412b905/ide/static/js/pane.js#L457-L466

The 'Wrapper' section in the pane.js file also has the same id, so the padding was being applied to that too. So I changed the id of the div in modelZoo.js, and the corresponding property name in zoo.css from 'wrapper' to 'content-wrapper'

